PR #6, opened 2018-03-06
Simplify user by adding a user session

deviantintegral:
https://codeclimate.com/github/Lullabot/mpx-php/src/User.php calls out the number of constructor parameters, which makes sense. My goals for this PR are to:

- [x] Split User into three classes. User will represent just the data for a user. UserSession will represent an active login session. TokenCachePool will hold all user tokens.
- [x] Prepare for making UserSession the main dependency for all authenticated requests. For example, instead of passing a Client through to a feed to fetch media items, we pass the user session. The main advantage of this is that if a request 403's due to an expired or cleared token, we can automatically retry by generating a new token, saving callers that trouble and error handling.
- [x] Remove optional / null method parameters and returns (most notably the possibly-null-return with a user token)
- [x] ~~The ID currently returned by `protected function getSelfId() {` is included in the token response. Since selfId is itself an authenticated request, we can remove this and store the id from the original session.~~ I decided not to keep this method as the D7 code doesn't use this API AFAICT. We can add it back later if it's useful.


src/TokenCachePool.php
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * Generate a cache key for a token, limiting key length.
+     *
+     * @param \Mpx\User $user
+     *                        The user to generate the cache key for.
+     *
+     * @return string
+     *                The cache key for the user.
+     */
+    private function cacheKey(User $user): string
```
davereid: Is this tested with PHP 7.0? Typehinting is only valid in 7.1, correct? Indents also seem off in the phpdoc.

deviantintegral: Yep, the [php7.0](https://circleci.com/gh/Lullabot/mpx-php/59?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) job passed.

PHP 7.0 supports both return type declarations and scalars in type hints: http://php.net/manual/en/migration70.new-features.php

deviantintegral: The indents actually are valid according the Symfony spec, but I've been going through and inlining them. I'll get this one done!

PR #14, opened 2018-03-06
Store created times instead of expiration times in the token

deviantintegral:
A minor cleanup; it makes more sense to me to store the time a token was created, instead of the result of a calculation.

Depends on #13 .


juanolalla:
:+1:

Issue #16, opened 2018-03-07
Support retries when an authenticated request fails


deviantintegral:
We should proxy all authenticated requests through `UserSession`. That way, if a 401 is returned, we can automatically renew the authentication token without API consumers having to manage that on their own.

Issue #9, opened 2018-03-06
Make logging configuration explicit


deviantintegral:
Currently, we have a `HasLoggerTrait` that defaults to a null logger. This means that if calling code uses a class that imports the trait, by default logging will be ignored - and it won't be obvious that logging is even a function that's available.

Instead, let's pull logging out to the constructor, and document that NullLogger is the best way to kill logs.

Also, let's give calling code the ability to log if possible, instead of doing it in the library. That means systems like Drupal can have proper translation and placeholders for user-facing messages. Let's keep logging for places where we do something informational that doesn't correspond to a return (like refreshing an authentication token inline with an API request).

PR #17, opened 2018-03-07
Make logger a required constructor parameter #9

deviantintegral:
This ensures that API consumers are aware of the logging functionality (which will be more important when we implement #16 ).


src/UserSession.php
```diff
@@ -70,6 +81,14 @@ public function acquireToken($duration = null): Token
         } catch (\RuntimeException $e) {
             // @todo This catch should be a more specific exception.
             $token = $this->signIn($duration);
+            $this->logger->info(
+                'Refreshed token with a new mpx token {token} for user {username} that expires on {date}.',
```
juanolalla, line 85: It's not clear to me why an exception thrown after storing in $token the result of $this->tokenCachePool->getToken() means that a token has been refreshed with a new one. And if that exception is thrown inside getToken(), shouldn't we implement the catch and the logger dependency in TockenCachePool::getToken()?

juanolalla (approved):
I'm just asking for clarification, everything looks good. I would also love an explanation about how that test is working, and why are you even using a regular expression on the mock callback. I guess we should hangout sometime :smile: 

Nice work!

src/UserSession.php
deviantintegral: If that exception is thrown, then the first `$token` will never be set. The class responsibilities look like:

* `UserSession`: MPX API calls - signing in and signing out.
* `TokenCachePool`: Caching MPX tokens.

If we push fetching the new token into `TokenCachePool`, then callers won't have a choice to decide how the retry logic should work without some more significant class coupling. Also, it would mean we would have to pass the MPX client and the MPX user into the TokenCachePool's constructor, which would increase it's complexity.

Is that clearer? Anything in the code that would help explain it?

PR #19, opened 2018-03-07
Use a middleware for exception handling

### LINKED ISSUE #3: Extend Guzzle client or wrap around?
There are two options when creating a client object:

1. Have my class extend GuzzleHttp\Client and make calls to `$this->{method}`
2. Have my class take a GuzzleHttp\Client object as a constructor parameter and make calls to `$this->client->{method}`

I tend to prefer option 1 since it means calling code doesn't have to deal with two Client objects.

### LINKED ISSUE #7: Return a request object instead of an array in Client
`\Mpx\Client::request` currently suppresses HTTP exceptions and returns an array. The MPX documentation says we can rely on response codes, so we should use them. Also, returning a response lets callers inspect headers (like cache control).

I'm thinking it makes the most sense to keep this functionality, but as other methods so callers aren't forced into this model.

deviantintegral:
A few of our issues were somewhat coupled with each other, where it made sense to solve them all at once. Here's the summary:

# MPX Client decoration

This PR resolves #3 by decorating the underlying Guzzle client. To use the MPX client, you first call `getDefaultConfiguration()`, use that to create the normal Guzzle HTTP client, and then pass that in to MPX. This lets other systems (like Drupal) merge in their own default configuration (like proxies, user agents, and timeouts).

# Returning a request object

This PR resolves #7 by adding a [Guzzle middleware](http://docs.guzzlephp.org/en/stable/handlers-and-middleware.html) that checks MPX responses for errors.

TODOs:

- [x] Fix that if MPX actually returns an HTTP error that we handle that properly.


juanolalla:
What a lot of work. I don't understand everything but it looks nice so far :smile: 

This got conflicts with a previous merged branch (the logger stuff), please resolve conflicts.

deviantintegral:
Thanks @juanolalla - feel free to poke if you want to hop on a call to ask about anything you'd like to learn about.

PR #22, opened 2018-03-09
Support automatic retry for authenticated requests #16

deviantintegral:
This PR solves #16 by automatically retrying a request if it fails with a 401, which indicates that the token has expired. While we automatically expire tokens locally, it's possible for them to be deleted server side or through a signout call.

There's a little bit of code duplication in that we want to be sure we call the appropriate underlying client method. Technically we could maintain the API contract even if we convert a blocking request into a promise, and then wait ourselves, but that would mean that the `request()` and `send()` methods in the HTTP client wouldn't be used, even when the calling code is completely blocking.

### DIFF
--- a/src/UserSession.php
+++ b/src/UserSession.php
@@ -224,11 +224,11 @@ class UserSession implements ClientInterface
      */
     private function mergeAuth(array $options, bool $reset = false): array
     {
-        $options += [
-            'auth' => [
-                $this->user->getUsername(),
-                $this->acquireToken(null, $reset)->getValue(),
-            ],
+        if (!isset($options['query'])) {
+            $options['query'] = [];
+        }
+        $options['query'] += [
+            'token' => $this->acquireToken(null, $reset)->getValue(),
         ];
 
         return $options;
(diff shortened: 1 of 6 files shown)


src/UserSession.php
```diff
@@ -171,4 +246,141 @@ private function tokenFromResponse(array $data): Token
...
+     * @param string                                $method  HTTP method
+     * @param string|\Psr\Http\Message\UriInterface $uri     URI object or string.
+     * @param array                                 $options Request options to apply.
+     *
+     * @return \GuzzleHttp\Promise\PromiseInterface|\Psr\Http\Message\RequestInterface
+     */
+    private function sendAsyncWithRetry(RequestInterface $request, array $options)
+    {
+        /** @var \GuzzleHttp\Promise\Promise $promise */
+        $promise = new Promise(function () use (&$promise, $request, $options) {
```
deviantintegral: This was easily the most complex part of the code to figure out, as the initial implementation was actually blocking. Hopefully the doc comment above makes this code a little clearer.

The `&$promise` was new to me - it lets the anonymous function access $promise, even though it technically hasn't been defined yet when the function is being created. Neat!

PR #34, opened 2018-03-13
Include the HTTP response code in exception messages

deviantintegral:


deviantintegral:
Was merged by mistake.

PR #38, opened 2018-03-15
Feature branch: Loading accounts and media

deviantintegral:
This PR adds an account object for loading MPX accounts. As there will be a decent amount of refactoring to integrate media and service resolution APIs, I'm going to use this as a long-lived branch until I'm comfortable writing tests.


src/DataService/Account/Account.php
```diff
@@ -0,0 +1,499 @@
...
+            $normalizers = [new ObjectNormalizer()];
+
+            $serializer = new Serializer($normalizers, $encoders);
+            $results = \GuzzleHttp\json_decode($response->getBody(), true);
+            $accounts = [];
+            foreach ($results['entries'] as $entry) {
+                // @todo Implement a QueryResult object that is returned, that in turn deserializes each object?
+                $encoded = \GuzzleHttp\json_encode($entry);
+                /** @var self $account */
+                $account = $serializer->deserialize($encoded, self::class, 'json', [
```
juanolalla: I'm guessing this deserialize() method is responsible for mapping the properties of each response 'entries' item to the properties of this Account class, is that right?

Is there some documentation link about what The Platform should be returning? I haven't found the Account object there.

deviantintegral: That's right, it's [the ObjectNormalizer](https://symfony.com/doc/3.4/components/serializer.html) above that looks for public properties and set() methods, calling them automatically.

The account object is at https://docs.theplatform.com/help/wsf-account-endpoint

juanolalla: It would be nice to add that link as @see before the class definition

I've seen loadAllAccounts() has been removed here anyway: https://github.com/Lullabot/mpx-php/pull/43/files. It would had been better to review all these pull request starting from newer...

deviantintegral:
Once #48 is merged to this branch, we should have enough test coverage to be able to merge to master.

PR #39, opened 2018-03-16
Add an initial ResolveAllUrls implementation

deviantintegral:
This depends on #38.

TODOs:

- [x] More class documentation
- [x] Clean up the `InvalidArgumentException`
- [x] Tests

### DIFF
--- a/src/Service/AccessManagement/ResolveAllUrls.php
+++ b/src/Service/AccessManagement/ResolveAllUrls.php
@@ -48,7 +48,7 @@ class ResolveAllUrls
     public function __construct(string $service, array $data)
     {
         if (!isset($data['resolveAllUrlsResponse'])) {
-            throw new \InvalidArgumentException();
+            throw new \InvalidArgumentException('Data does not contain a resolveAllUrlsResponse key and does not appear to be an MPX response.');
         }
 
         $this->resolved = $data['resolveAllUrlsResponse'];
--- a/src/Service/AccessManagement/ResolveAllUrls.php
+++ b/src/Service/AccessManagement/ResolveAllUrls.php
@@ -5,16 +5,28 @@ namespace Lullabot\Mpx\Service\AccessManagement;
 use Lullabot\Mpx\Service\IdentityManagement\UserSession;
 use Psr\Http\Message\ResponseInterface;
 
+/**
+ * Resolve all URLs for a given service.
+ *
+ * @see https://docs.theplatform.com/help/wsf-resolveallurls-method
+ */
 class ResolveAllUrls
 {
+    /**
+     * While this method is not a data service, it still has a schema version.
+     */
     const SCHEMA_VERSION = '1.0';
 
+    /**
+     * The URL of the method. Note that we hardcode the whole path as this is
+     * where other services are bootstrapped from.
+     */
     const RESOLVE_ALL_URLS_URL = 'https://access.auth.theplatform.com/web/Registry/resolveAllUrls';
 
     /**
      * An array of resolved URLs for the service.
      *
-     * @var array
+     * @var string[]
      */
     protected $resolved;
 
@@ -30,8 +42,8 @@ class ResolveAllUrls
      *
      * @todo Is there value in storing Responses in all classes?
      *
-     * @param string $service
-     * @param array  $data
+     * @param string $service The service that was queried, such as 'Access Data Service'.
+     * @param array  $data    The MPX response.
      */
     public function __construct(string $service, array $data)
     {
@@ -43,6 +55,14 @@ class ResolveAllUrls
         $this->service = $service;
     }
 
+    /**
+     * Fetch URLs and return the response.
+     *
+     * @param \Lullabot\Mpx\Service\IdentityManagement\UserSession $userSession The authenticated session to use when querying.
+     * @param string                                               $service     The service to find URLs for, such as 'Media Data Service'.
+     *
+     * @return \GuzzleHttp\Promise\PromiseInterface A promise to return a new ResolveAllUrls class.
+     */
     public static function load(UserSession $userSession, string $service)
     {
         $options = [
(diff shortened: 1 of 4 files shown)


juanolalla:
I don't know what is this class about, looking forward for the class documentation

deviantintegral:
https://docs.theplatform.com/help/wsf-registry-endpoint-access-management is a good starting point.

deviantintegral:
@juanolalla this is ready for review whenever you are.

src/Service/AccessManagement/ResolveAllUrls.php
```diff
@@ -0,0 +1,99 @@
...
+     * @param string $service The service that was queried, such as 'Access Data Service'.
+     * @param array  $data    The MPX response.
+     */
+    public function __construct(string $service, array $data)
+    {
+        if (!isset($data['resolveAllUrlsResponse'])) {
+            throw new \InvalidArgumentException('Data does not contain a resolveAllUrlsResponse key and does not appear to be an MPX response.');
+        }
+
+        $this->resolved = $data['resolveAllUrlsResponse'];
```
juanolalla, line 54: I'm not sure if this property makes sense since it's always going to be true.

tests/src/Unit/Service/AccessManagement/ResolveAllUrlsTest.php
```diff
@@ -0,0 +1,58 @@
...
+            new JsonResponse(200, [], 'signin-success.json'),
+            new JsonResponse(200, [], 'resolveAllUrls.json'),
+        ]);
+        $user = new User('USER-NAME', 'correct-password');
+        $tokenCachePool = new TokenCachePool(new ArrayCachePool());
+        $session = new UserSession($client, $user, $tokenCachePool, new NullLogger());
+        /** @var \Lullabot\Mpx\Service\AccessManagement\ResolveAllUrls $r */
+        $r = ResolveAllUrls::load($session, 'Media Data Service')->wait();
+        $this->assertEquals('Media Data Service', $r->getService());
+        $this->assertEquals(['http://data.media.theplatform.com/media'], $r->getResolved());
```
juanolalla, line 44: if the previous assert was correct, we know the response is resolved, otherwise a exception would had been thrown. I think getResolved() is redundant and we can get rid of it.

src/Service/AccessManagement/ResolveAllUrls.php
```diff
@@ -0,0 +1,99 @@
...
+    {
+        $options = [
+            'query' => [
+                'schema' => static::SCHEMA_VERSION,
+                '_service' => $service,
+            ],
+        ];
+
+        return $userSession->requestAsync('GET', self::RESOLVE_ALL_URLS_URL, $options)->then(function (ResponseInterface $response) use ($service) {
+            return new static($service, \GuzzleHttp\json_decode($response->getBody(), true));
```
juanolalla, line 76: Is it possible that $response->getBody() doesn't return a 'resolveAllUrlsResponse' in the decoded array? In that case, it would be nice to have a test for that, executing ::load()->wait() and then expecting the exception.

juanolalla:
Hey @deviantintegral, why didn't you answer to my questions here?

juanolalla (commented):
So sorry I didn't submit this

src/Service/AccessManagement/ResolveAllUrls.php
deviantintegral: true? It's an array of resolved URLs such as:

```
{
    "resolveAllUrlsResponse": [
        "http://mps.theplatform.com",
        "http://data.media.theplatform.com/media"
    ]
}
```

tests/src/Unit/Service/AccessManagement/ResolveAllUrlsTest.php
deviantintegral: Great idea. I've removed the method in 5b100fd .

src/Service/AccessManagement/ResolveAllUrls.php
deviantintegral: Did you see the [check in the constructor](https://github.com/Lullabot/mpx-php/blob/7056be96664d6e8efdf8765fabdc14ab03b87e58/src/Service/AccessManagement/ResolveAllUrls.php#L50) and `\Lullabot\Mpx\Tests\Unit\Service\AccessManagement\ResolveAllUrlsTest::testInvalidData`?

src/Service/AccessManagement/ResolveAllUrls.php
juanolalla: Sorry, I thought you were storing that just for the sake of a boolean check, meaning: is this resolved? Not actually because we might need to retrieve the result stored there. The word confused me.

src/Service/AccessManagement/ResolveAllUrls.php
juanolalla: Yes, I saw that, that is testing that if you manually instantiated the class with a malformed array it would throw an exception. I just thought that would be better if we could test that by calling load() with an URL that doesn't result a response with 'resolveAllUrlsResponse'. But maybe it's unnecessary and redundant as it is the code right now.

src/Service/AccessManagement/ResolveAllUrls.php
deviantintegral: Ah, got it! I will rename the method and variable to `getResolvedUrls` tomorrow.

src/Service/AccessManagement/ResolveAllUrls.php
deviantintegral: I see what you're thinking. I don't think we need the additional test, as the `load()` method doesn't actually do any validation and must call `__construct` to satisfy the return value. Does that work for you?

juanolalla: I think you're right, we are testing the same already.

src/Service/AccessManagement/ResolveAllUrls.php
juanolalla: oh! that's perfect! thanks

PR #40, opened 2018-03-16
Initial media data object class

deviantintegral:
This depends on #39 and will resolve #27 .

This is mostly generated code with getters and setters for each property. There intentionally isn't a constructor; Symfony's serializer doesn't allow constructors with parameters for the default `ObjectNormalizer`. We may add one later if we decide to implement a custom normalizer, but for now it is aiming for the simple path.

PR #41, opened 2018-03-16
Add an annotation for MPX data services

deviantintegral:
Depends on #40 .

It took me a while to get to this implementation, so let me explain.

My first attempt locally was to have static methods like `loadById()` on each data class like Media and Account. However, we know statics are harder to test later. Also, much of the point of a static is to call a constructor, which we wouldn't have using the default Symfony serializer. Also, the static methods needed a lot of metadata - the schema version, the service name (like `Media Data Service`, the internal object path (`/data/Media`), and so on. Then, I realized that with those variables, the code was nearly identical between multiple types of MPX data - in other words, the API is well-structured enough that a simple description of the data gives us enough information to load it. Neat!

Many of the APIs needed to load required many parameters; the active session, the account context, the type of data being loaded, the destination class, and so on. 4 or 5 parameters on the static methods told me it wouldn't scale for any MPX data that ended up being just slightly more complicated. So, I pulled the static methods out to a separate "Description" object and interface.

From that, I realized all of the metadata we needed was truly "static" - it would change if the MPX docs and APIs updated, but would never change at runtime. With slight horror, I realized that annotations were perfect - because if MPX added a new field, you knew to update the annotation schema version, add the field, and everything else would just work.

The other advantage of this approach is that developers could implement parallel classes for different schema versions. Perhaps a consuming app is _very_ dependent on Media schema 1.9. That class instance could live on forever "in ice", and the annotation discovery could return classes-per-schema. It's not implemented in this PR, but it's easy to add and probably worth doing before a 1.0 tag.

I will add notes below. Assuming this all makes sense, once we get everything merged into the (poorly named) `account-object` branch I will write tests there before merging to master.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,5 +34,8 @@
       "Lullabot\\Mpx\\Tests\\": "tests/src/",
       "Lullabot\\Mpx\\Command\\": "command/src/"
     }
+  },
+  "extra": {
+    "composer-version": "^1.6.3"
   }
 }
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,8 @@
     "psr/log": "^1.0",
     "cache/cache": "^1.0",
     "symfony/serializer": "^3.4",
-    "symfony/property-access": "^3.4"
+    "symfony/property-access": "^3.4",
+    "doctrine/annotations": "^1.4"
   },
   "require-dev": {
     "phpunit/phpunit": "^6.0",
--- a/src/CreateKeyInterface.php
+++ b/src/CreateKeyInterface.php
@@ -11,6 +11,8 @@ use Psr\Http\Message\UriInterface;
  * create request, instead of creating a new object.
  *
  * @see https://docs.theplatform.com/help/wsf-creating-or-updating-data-objects-using-create-keys
+ *
+ * @todo Move to the annotation and convert to CustomKeyInterface.
  */
 interface CreateKeyInterface
 {
--- a/src/DataService/Annotation/DataService.php
+++ b/src/DataService/Annotation/DataService.php
@@ -1,0 +1,113 @@
+<?php
+
+namespace Lullabot\Mpx\DataService\Annotation;
+
+/**
+ * @Annotation
+ *
+ * Each MPX data service exposes one or more objects. We require annotations
+ * on object implementations so calling code can discover what services are
+ * currently implemented. In general, there should only be one implementation
+ * of a given (path, service, schema) triple.
+ *
+ * @todo Use symfony validation to assert requirements.
+ * @todo Mark the required values with @Required
+ * @todo Rename to DataServiceObject
+ * @todo Can we infer /data/ from path and assume it's always consistent?
+ *
+ * @see http://docs.doctrine-project.org/projects/doctrine-common/en/latest/reference/annotations.html
+ */
+class DataService
+{
+    /**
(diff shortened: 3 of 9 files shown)


composer.json
```diff
@@ -34,5 +35,8 @@
       "Lullabot\\Mpx\\Tests\\": "tests/src/",
       "Lullabot\\Mpx\\Command\\": "command/src/"
     }
+  },
+  "extra": {
+    "composer-version": "^1.6.3"
```
deviantintegral, line 40: This is for `composer-gavel` support and will be ignored if the plugin isn't installed.

composer.json
```diff
@@ -15,7 +15,8 @@
     "psr/log": "^1.0",
     "cache/cache": "^1.0",
     "symfony/serializer": "^3.4",
-    "symfony/property-access": "^3.4"
+    "symfony/property-access": "^3.4",
+    "doctrine/annotations": "^1.4"
```
deviantintegral, line 19: This is specifically 1.4 so we preserve PHP 7.0 compatibility.

src/CreateKeyInterface.php
```diff
@@ -11,6 +11,8 @@
  * create request, instead of creating a new object.
  *
  * @see https://docs.theplatform.com/help/wsf-creating-or-updating-data-objects-using-create-keys
+ *
+ * @todo Move to the annotation and convert to CustomKeyInterface.
```
deviantintegral, line 15: I may remove this entirely, as two of the three methods are covered by the annotation. However, I'm unsure yet if developers will be able to statically define custom keys in a configuration somewhere, or if it needs to be runtime code.

src/DataService/Annotation/DataService.php
```diff
@@ -0,0 +1,113 @@
+<?php
+
+namespace Lullabot\Mpx\DataService\Annotation;
+
+/**
+ * @Annotation
+ *
+ * Each MPX data service exposes one or more objects. We require annotations
```
deviantintegral, line 8: I'm not a huge fan of 'DataService', but it's what MPX calls each service in their docs, and `MpxDataService` seemed redundant with the namespace.

src/DataService/DataServiceDiscovery.php
```diff
@@ -0,0 +1,106 @@
...
+
+namespace Lullabot\Mpx\DataService;
+
+use Doctrine\Common\Annotations\Reader;
+use Symfony\Component\Finder\Finder;
+
+/**
+ * Class to discover Data Service implementations.
+ */
+class DataServiceDiscovery
```
deviantintegral, line 11: My guess is that a Drupal or other framework implementation would replace this class, but we still need it for our own testing, and it's useful as a guide.

src/DataService/DataServiceManager.php
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * Register our annotations, relative to this file.
+     *
+     * @return static
+     */
+    public static function basicDiscovery()
+    {
+        // @todo Check Drupal core for other tags to ignore?
+        AnnotationReader::addGlobalIgnoredName('class');
+        AnnotationRegistry::registerFile(__DIR__.'/Annotation/DataService.php');
```
deviantintegral, line 29: This line is critical, otherwise `Access` is loaded first, which uses the DataService annotation, and you end up with a class not found fatal.

PR #43, opened 2018-03-16
Implement annotations for Media and Accounts

deviantintegral:
Depends on #42 .

This PR adds the proper annotations so we can load both accounts and media.

Testing instructions:

* You need your MPX username and password.
* You need an account ID.
* You need a media ID.

Place the following in the root of the repository as a PHP file like `test.php`:

```php
use Cache\Adapter\PHPArray\ArrayCachePool;
use Lullabot\Mpx\Client;
use Lullabot\Mpx\DataService\DataObjectFactory;
use Lullabot\Mpx\DataService\DataServiceManager;
use Lullabot\Mpx\Service\IdentityManagement\UserSession;
use Lullabot\Mpx\TokenCachePool;
use Lullabot\Mpx\User;
use Psr\Log\NullLogger;

require_once './vendor/autoload.php';

// Create a new MPX client with the default configuration.
$defaults = Client::getDefaultConfiguration();
$client = new Client(new GuzzleHttp\Client($defaults));

// Replace your username and password here. The username must begin with `mpx/`.
$user = new User('mpx/YOU@EXAMPLE.COM', 'secret');
$tokenCachePool = new TokenCachePool(new ArrayCachePool());
$session = new UserSession($client, $user, $tokenCachePool, new NullLogger());

// This registers the annotation loader. I suppose we could call this from DataObjectFactory to make it automatic?
$dataServiceManager = DataServiceManager::basicDiscovery();

$accountFactory = new DataObjectFactory($dataServiceManager, $session, 'Access Data Service', '/data/Account');

// Replace the ID with the account ID to load.
$account = $accountFactory->load(12345)
    ->wait();
print "The loaded account is:\n";
var_dump($account);

$mediaFactory = new DataObjectFactory($dataServiceManager, $session, 'Media Data Service', '/data/Media');

// Replace the ID to the media item to load. You can find it under "History -> ID" in the MPX console.
$media = $mediaFactory->load(12345, $account)
    ->wait();
print "The loaded media is:\n";
var_dump($media);
```

### DIFF
--- a/src/DataService/Annotation/DataService.php
+++ b/src/DataService/Annotation/DataService.php
@@ -1,0 +1,113 @@
+<?php
+
+namespace Lullabot\Mpx\DataService\Annotation;
+
+/**
+ * @Annotation
+ *
+ * Each MPX data service exposes one or more objects. We require annotations
+ * on object implementations so calling code can discover what services are
+ * currently implemented. In general, there should only be one implementation
+ * of a given (path, service, schema) triple.
+ *
+ * @todo Use symfony validation to assert requirements.
+ * @todo Mark the required values with @Required
+ * @todo Rename to DataServiceObject
+ * @todo Can we infer /data/ from path and assume it's always consistent?
+ *
+ * @see http://docs.doctrine-project.org/projects/doctrine-common/en/latest/reference/annotations.html
+ */
+class DataService
+{
+    /**
+     * The name of the service, such as 'Media Data Service'.
+     *
+     * @var string
+     */
+    public $service;
+
+    /**
+     * The relative path of the data service, such as '/data/Media'.
+     *
+     * @var string
+     */
+    public $path;
+
+    /**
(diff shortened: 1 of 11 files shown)


src/DataService/Annotation/DataService.php
```diff
@@ -0,0 +1,113 @@
...
+ * on object implementations so calling code can discover what services are
+ * currently implemented. In general, there should only be one implementation
+ * of a given (path, service, schema) triple.
+ *
+ * @todo Use symfony validation to assert requirements.
+ * @todo Mark the required values with @Required
+ * @todo Rename to DataServiceObject
+ * @todo Can we infer /data/ from path and assume it's always consistent?
+ *
+ * @see http://docs.doctrine-project.org/projects/doctrine-common/en/latest/reference/annotations.html
```
juanolalla, line 18: Thanks for this link! This is just what I need to read right now to better understand all this approach, which sounds nice so far by the way.

PR #50, opened 2018-03-20
Add Overview of main classes to README.md

juanolalla:
This is something I wrote while I was learning about our code, I thought it might be a good idea to update the README.md with it so a new developer can get a quick overview before getting into the code more deeply.

Please review it so to ensure it's accurate and well written.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,3 +19,22 @@ API request that resulted in a `401`.
 If your application does not wish to log these actions at all, use
 `\Psr\Log\NullLogger` for any constructors that require a
 `\Psr\Log\LoggerInterface`.
+
+
+## Overview of main classes
+
+### Client
+MPX API implementation of Guzzle ClientInterface. As a Client it doesn’t do anything extra but suppress errors to force a returning HTTP 200.
+It also handles XML from responses
+
+### UserSession
+A wrapper around Client, implementing all the stuff from Guzzle ClientInterface, and adding sign and singout functionalities.
+
+### User
+An MPX user. Just username and password getters.
+
+### Token
+MPX authentication token that is returned by the platform after [sign in](https://docs.theplatform.com/help/wsf-signin-method)
+
+### TokenCachePool
+Cache of user authentication tokens. This class is a wrapper around a \Psr\Cache\CacheItemPoolInterface object
--- a/README.md
+++ b/README.md
@@ -23,18 +23,19 @@ If your application does not wish to log these actions at all, use
 
 ## Overview of main classes
 
-### Client
-MPX API implementation of Guzzle ClientInterface. As a Client it doesn’t do anything extra but suppress errors to force a returning HTTP 200.
-It also handles XML from responses
+### Lullabot\Mpx\Client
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,3 +19,22 @@ API request that resulted in a `401`.
 If your application does not wish to log these actions at all, use
 `\Psr\Log\NullLogger` for any constructors that require a
 `\Psr\Log\LoggerInterface`.
+
+
+## Overview of main classes
+
+### Client
```
deviantintegral: Let's make these the fully-qualified class paths?

README.md
```diff
@@ -19,3 +19,22 @@ API request that resulted in a `401`.
...
+MPX API implementation of Guzzle ClientInterface. As a Client it doesn’t do anything extra but suppress errors to force a returning HTTP 200.
```
deviantintegral: Let's link to ClientInterface over in the Guzzle repository?

`suppress errors to force a returning HTTP 200.` => `suppress errors to force a returning HTTP 200 for XML responses.`

README.md
```diff
@@ -19,3 +19,22 @@ API request that resulted in a `401`.
...
+It also handles XML from responses
```
deviantintegral: Missing period.

README.md
```diff
@@ -19,3 +19,22 @@ API request that resulted in a `401`.
...
+
+### UserSession
+A wrapper around Client, implementing all the stuff from Guzzle ClientInterface, and adding sign and singout functionalities.
```
deviantintegral: Let's reverse the order of this - the important part is the session, not the method implementations.

> Manages authenticated sessions and proxies API calls with a ClientInterface implementation, automatically refreshing expired API tokens as required.

README.md
```diff
@@ -19,3 +19,22 @@ API request that resulted in a `401`.
...
+
+### User
+An MPX user. Just username and password getters.
+
+### Token
+MPX authentication token that is returned by the platform after [sign in](https://docs.theplatform.com/help/wsf-signin-method)
```
deviantintegral: Period!

README.md
```diff
@@ -19,3 +19,22 @@ API request that resulted in a `401`.
...
+
+### TokenCachePool
+Cache of user authentication tokens. This class is a wrapper around a \Psr\Cache\CacheItemPoolInterface object
```
deviantintegral: Period!

It's not a wrapper, but an implementation of? A wrapper typically implies [decoration](https://www.lullabot.com/articles/get-a-decorator-for-your-drupal-home), whereas an implementation fills in the requirements from one or more interfaces.

deviantintegral (commented):
Huh, odd that code coverage changed on this PR. I will take a look.

I like where this is going! Later, before a 1.0 release, I'd like us to have some concrete examples in the README of using the library.

deviantintegral:
The PR branch was against an older master commit. Merging master cleared the code coverage report.

README.md
juanolalla: Actually is not an implementation of any interface, it's using a CacheItemPoolInterface object to add additional features, very similar to what a decorator does, although I used the term wrapper without knowing exactly that connotation of the word, just because it seems it's wrapping the object and adding some extra stuff.

deviantintegral (approved):

PR #52, opened 2018-03-20
Initial notification listener

### LINKED ISSUE #23: Support media polling with notifications
* Double check to make sure that there isn't a webook / push API we can use instead.
* Implement a Media Notification object that takes a UserSession and calls `https://read.data.media.theplatform.com/media/notify`.
* Investigate returned data to see if we should have a single Media object for both "data" and "notifications.

deviantintegral:
This will address part of #23 .

Also, we are currently affected by https://github.com/guzzle/guzzle/pull/2001

### DIFF
--- a/src/DataService/DataObjectFactory.php
+++ b/src/DataService/DataObjectFactory.php
@@ -2,6 +2,7 @@
 
 namespace Lullabot\Mpx\DataService;
 
+use GuzzleHttp\Promise\PromiseInterface;
 use Lullabot\Mpx\DataService\Access\Account;
 use Lullabot\Mpx\Service\AccessManagement\ResolveDomain;
 use Lullabot\Mpx\Service\IdentityManagement\UserSession;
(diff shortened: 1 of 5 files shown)


src/DataService/DataObjectFactory.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Lullabot\Mpx\DataService;
 
+use GuzzleHttp\Promise\PromiseInterface;
```
deviantintegral, line 5: You can ignore the changes in this file as I am already improving them in a later PR.

src/DataService/NotificationListener.php
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * The user session to use for authenticated requests.
+     *
+     * @var UserSession
+     */
+    protected $session;
+
+    /**
+     * The name of the service to listen to notifications on, such as 'Media Data Service.'.
```
juanolalla: Is the ending dot included in the name of the service? You've written 'Media Data Service.'

src/DataService/NotificationListener.php
```diff
@@ -0,0 +1,132 @@
...
+     * @var string
+     */
+    private $clientId;
+
+    /**
+     * NotificationListener constructor.
+     *
+     * @param string      $clientId A string to identify this client in debugging.
+     * @param UserSession $session  The user session to use for authenticated requests.
+     * @param string      $service  The name of the service to listen to notifications on, such as 'Media Data Service.'
```
juanolalla: I guess the dot is needed, since you included it here again

juanolalla (commented):
Just a question before merging

src/DataService/NotificationListener.php
juanolalla: Now I know it's not, because I've seen this service name in a few places more :)

src/DataService/NotificationListener.php
deviantintegral: No, it's not, good catch!

PR #53, opened 2018-03-20
Support read-only MPX endpoints

### LINKED ISSUE #49: Support read-only data service instances
> "If you only need read access to the the data, you can get better performance by using a read-only instance of the endpoint."
  -- [A Helpful Technical Writer](https://docs.theplatform.com/help/services)

Let's add support for this. It should be close to automatic, as each service simply has 'read-only' appended to it's name.

deviantintegral:
This will resolve #49 by adding an additional optional parameter to `load()`:

For example:

`$mediaFactory->load(1234, $account, true)`

Forces the read-only endpoint. Omitting the last parameter will query the read-write endpoint.


src/DataService/Access/Account.php
```diff
@@ -11,8 +11,7 @@
  *   service="Access Data Service",
  *   baseUri="https://access.auth.theplatform.com/data/Account",
  *   schemaVersion="1.0",
- *   path="/data/Account",
- *   hasAccountContext=false
```
deviantintegral, line 15: I read in their docs that _all_ data services support account contexts, so there's no need for it in the annotation.

src/DataService/Annotation/DataService.php
```diff
@@ -59,10 +50,16 @@ class DataService
     /**
      * Return the service of the data object, such as 'Media Data Service'.
      *
-     * @return string
+     * @param bool $readonly (optional) Set to true to return the read-only name of the service.
+     *
+     * @return string The name of the service.
      */
-    public function getService(): string
+    public function getService($readonly = false): string
```
deviantintegral, line 57: This is a boolean so if later we find that appending a string doesn't work that we can have whatever logic we need (such as pulling it from an annotation property).

src/DataService/DataObjectFactory.php
```diff
@@ -38,48 +38,33 @@ class DataObjectFactory
...
-     * @todo Merge manager and service parameters?
      * @todo Inject the resolveDomain() instead of constructing?
      *
-     * @param \Lullabot\Mpx\DataService\DataServiceManager         $manager
+     * @param array                                                $description The array describing the destination class for this factory.
      * @param \Lullabot\Mpx\Service\IdentityManagement\UserSession $userSession
-     * @param string                                               $service
      */
-    public function __construct(DataServiceManager $manager, UserSession $userSession, string $service, string $path)
+    public function __construct(array $description, UserSession $userSession)
```
deviantintegral, line 49: I think I might change the description array to a real class later.

juanolalla (approved):

juanolalla:
Not sure if we are done with #49 and must be closed

deviantintegral:
Yes, we are, thanks!

Issue #28, opened 2018-03-11
Investigate merging Client and UserSession


deviantintegral:
~~There's no namespace separation for most of our classes. We should find some commonalities and fix that.~~

> So my proposal is that UserSession would not implement the ClientInterface, but keeping it just for the purpose of managing user session functionality, as sign in and sign out.

juanolalla:
It looks to me that it doesn’t make much sense to have two classes implementing Guzzle ClientInterface . The logic of the Client is implemented mainly in UserSession, except for that only line to force HTTP 200 in Client::request(). So the functionality Client is actually implementing to handle XML from a Response could be done without needing to implement the ClientInterface, and it could have a different name other than Client, more adjusted to its XML utilities purpose.

As the same time UserSession seems confusing. It’s like it’s implementing two different purposes, one is the Client we need, and the other is UserSession functionality, like sign in or sign out. It would be nice to split those in two classes, one being actually the Client, and the other being an UserSession which can be managed by the client as dependency if needed, as the use of acquireToken() in mergeAuth() (if we think this belongs more to UserSession than to the Client).

What do you think?

deviantintegral:
> except for that only line to force HTTP 200 in Client::request()

I should document this, but MPX actually forces this server-side as soon as you request JSON. I think the idea was that if you are requesting XML this would keep the behaviour consistent, but I'm not sure if we'll need XML at all. `cjson` works almost totally out of the box with Symfony's decoder (while the MPX XML uses namespaces, which is not supported by XMLEncoder).

> the other being an UserSession which can be managed by the client as dependency if needed

At one point locally, I had something like this. I thought it made the initial authentication workflow a little tricky. In this case, a user session would have to be optional (as there are various anonymous APIs in MPX). That would mean that if you created a client, and then it was passed through an application, different code in the app wouldn't know the client was associated with a session. For example, in Drupal:

```php
$client = $this->container->get('mpx.client');

....

function loadVideo(Client $client) {
  // We don't know if $client has a user session, so we have to manually check.
  if (!$client->getUserSession()) {
    // Our class now needs a dependency on usernames and passwords to somehow create the session.
  }
}
```

If all API calls are routed through UserSession, then we can typehint against that:

```
$session = $this->container->get('mpx.user_session');

....

function loadVideo(UserSession $session) {
  // We know that $session will let us make an authenticated call.
}
```

[...]

PR #54, opened 2018-03-20
Add a lock around acquiring a token

### LINKED ISSUE #24: Support locks when signing in a user
Imagine:

* A site is deployed and all caches are cleared
* The site is very successful, and gets a ton of traffic
* The single, shared, cached token expires
* All of those requests see an expired token, and try to fetch a new one at the same time.
* This makes site owners and thePlatform sad as everything goes down.

We can fix this by supporting a locking service on the UserSession class.

- [x] Find out if there's a similar interface we can rely on like the PSR logging and cache interfaces, but for a lock service.
- [x] If not, see if any library (symfony? doctrine?) has one we can use.
- [x] Find or implement an equivalent to the NullLogger for logging. Making this a hard requirement (versus a set method) will help ensure implementors think about this potential DDoS target.

deviantintegral:
This resolves #24.


deviantintegral:
@juanolalla codeclimate is reasonably flagging the 5-parameter constructor. Looking at the class, I'm thinking of trying to address #28 by moving the authentication methods like signIn, and their dependencies, to the User class. For example, `$user->signIn(...)` would return a UserSession. Does that seem like a good path to you?

src/Service/IdentityManagement/UserSession.php
```diff
@@ -443,4 +459,31 @@ private function outerPromise(PromiseInterface $inner): Promise
...
+     *
+     * @param int $duration (optional) The number of seconds that the sign-in token should be valid for.
+     *
+     * @return Token
+     */
+    protected function signInWithLock(int $duration = null): Token
+    {
+        $factory = new Factory($this->store);
+        $factory->setLogger($this->logger);
+        $lock = $factory->createLock($this->user->getUsername(), 10);
```
deviantintegral: This creates a lock that lasts for 10 seconds. Note that any lock prefixing would be handled by decorating the store class that is passed in. The lock will release once `$lock` goes out of scope.

juanolalla:
@deviantintegral At first look it seems reasonable to move the authentication methods to User. Once we do that, we should ask ourselves what is the purpose of Client and UserSession then. I would like to see only one class implementing the Guzzle ClientInterface if we don't really need two. That could be Client I think. And having UserSession for the rest of purposes left in the workflow.

deviantintegral:
@juanolalla , I like where this is going. Now of course `User` has many constructor parameters and I'd like to fix that.

I'd like to add documentation and one more class so that:

* `\Lullabot\Mpx\Client` is the "root" MPX client. It works with all types of requests, authenticated or anonymous.
* `\Lullabot\Mpx\User` goes back to just having a username and password.
* `\Lullabot\Mpx\Service\IdentityManagement\UserSession` is renamed to `AuthenticatedClient`, stays as is in the current PR. It exists because of the complex retry / promise logic we need.
* A new class, but named `UserSession`, is forked out from what's currently in User. It handles the the token fetch and caching logic.

Plus, I'd add docs to the readme and class headers detailing all of this. Thoughts?

juanolalla:
Well that sounds really good! Since there are clear reasons to have two clients, I'm glad you're changing the name and separating the other logic to a new UserSession class. I like classes to have a specific purpose, so that means having more classes when you realize the current ones are doing too much.

PR #58, opened 2018-03-21
Add an example of loading data to the README

deviantintegral:
This is copied from the PR description at #43 .


README.md
```diff
@@ -9,6 +9,52 @@
 
 `composer require lullabot/mpx-php`
 
+## Example
```
juanolalla, line 12: I think it would be nice to explain a little bit what is this example about, before all the following code

README.md
```diff
@@ -9,6 +9,52 @@
...
+$client = new Client(new GuzzleHttp\Client($defaults));
+
+// Replace your username and password here. The username must begin with `mpx/`.
+$user = new User('mpx/YOU@EXAMPLE.COM', 'secret');
+$tokenCachePool = new TokenCachePool(new ArrayCachePool());
+$session = new UserSession($client, $user, $tokenCachePool, new NullLogger());
+
+// This registers the annotation loader.
+$dataServiceManager = DataServiceManager::basicDiscovery();
+
```
juanolalla, line 43: Looking at all that above, from a developer experience perspective, I think we could simplify all those lines a lot. In this example, the data input is just for the user ('mpx/YOU@EXAMPLE.COM' and 'secret'). All the rest of lines are trivial initialization objects that we could save to developers if we implemented a class/method that receives the username and password and set all the rest of default stuff internally, giving us the session, dataServiceManager, and/or whatever the developer might need to continue.

deviantintegral: I added an issue for just this suggestion over at #69.

README.md
```diff
@@ -11,6 +11,9 @@
 
 ## Example
 
+Here is a complete example showing how to load an account and media items from
+MPX.
```
juanolalla: :+1: Perfect!

README.md
juanolalla: Great! :+1:

PR #60, opened 2018-03-21
WIP Byfield select

### LINKED ISSUE #56: Support general data service queries
Currently, we can only load data by it's ID. We should also support generating a query to use to load a subset of objects, such as "all media matching a title".

deviantintegral:
Will resolve #56 and hopefully others!


deviantintegral:
See #67

PR #67, opened 2018-03-26
Support byFields, ranges, sorts, and paging

### LINKED ISSUE #56: Support general data service queries

deviantintegral:
Replaces #60 and resolves #56.

This pull request adds support for:

* Querying MPX on "by" fields like "byTitle".
* Sorting and setting ranges on those results such as "order by id, objects 11-20".
* Deserializing an "object list" of results from an MPX query. In other words, if an ObjectList contains 100 media items, each item will be transparently deserialized when the response is handled. As a part of this, I added the CJsonEncoder class to handle null entries and date / URI normalizers, letting us remove the trait and json_decode / encode hacks we had previously.
* Automatically paging through multiple ObjectLists. Unfortunately, MPX doesn't provide "next" and "previous" relationships in responses, so we have to calculate them ourselves.
* Array / Iterator implementations for individual pages of results, and an iterator for automatically paging through each group of results.

For example, the following:

1. Filters media to the specified title (note MPX does exact matches only)
1. Returns results in groups of 100
1. Sorts groups by title
1. Prints the title from each result

```php
$fields = new \Lullabot\Mpx\DataService\ByFields();
$fields->addField('title', 'My Excellent Video');
$range = new \Lullabot\Mpx\DataService\Range();
$range->setStartIndex(1)
    ->setEndIndex(100);
$fields->setRange($range);

$sort = new \Lullabot\Mpx\DataService\Sort();
$sort->addSort('title');
$fields->setSort($sort);

// Results is an iterator, so we can foreach over it.
$results = $mediaFactory->select($fields, $account);
/** @var \Lullabot\Mpx\DataService\Media\Media $m */
foreach ($results as $m) {
    print $m->getTitle() . "\n";
}
```

Since titles are exact matches, the easiest way to test the paging is to lower the end index down to two or three, and remove the title filter.


src/DataService/ByFields.php
```diff
@@ -0,0 +1,116 @@
...
+     * The array of fields and their value to filter by.
+     *
+     * @var array
+     */
+    protected $fields = [];
+
+    /**
+     * The sort to apply to this filter.
+     *
+     * @var Sort
```
juanolalla, line 26: I guess you're using the class name because it belongs to the same namespace we are in. But that this have real implications apart from readibility? I mean, does PhpStorm understand that this variable is \Lullabot\Mpx\DataService\Sort without the use clause?

juanolalla (approved):

juanolalla:
Impressive work!

src/DataService/ByFields.php
deviantintegral: Yes, it does! I don't know when PHPStorm added support for that, but it now works fine as long as the class is imported with a use statement. It surprised me too! Apparently the Symfony standards want you to shorten classes in docs, while Drupal wants the full path.

preferences 2018-03-29 06-58-47

juanolalla: > it now works fine as long as the class is imported with a use statement

That's what I mean,  you're not using the use statement!

deviantintegral: Ah, in that case it works because Sort and ByFields are in the same namespace.

PR #68, opened 2018-03-28
Support media and non-media notifications

### LINKED ISSUE #23: Support media polling with notifications

deviantintegral:
Resolves #23.

This PR fully supports media object notifications, and in fact any type of object that supports them

Highlights include:

* I pulled the trigger and changed the `path` annotation to `objectType`. If we find a service that doesn't have `/data/` in it's path, we can always add a new annotation key for that case.
* A new `DiscoveredDataService` class replacing the class / annotation array.
* We now index data services also by schema version, so we can totally support multiple versions of an object in the same codebase.
* MPX highly recommends filtering to an object type for notifications. So for now, we force that by only ever listening for one object type per `NotificationListener` instance. If calling code needs to listen for multiple types of objects, they can call listen multiple times and wait on the promises.

### DIFF
--- a/src/DataService/DataServiceDiscovery.php
+++ b/src/DataService/DataServiceDiscovery.php
@@ -65,7 +65,7 @@ class DataServiceDiscovery
     /**
      * Returns all the data services.
      *
-     * @return DiscoveredDataService[] An array of all discovered data services.
+     * @return DiscoveredDataService[] An array of all discovered data services, indexed by service name, object type, and schema version.
      */
     public function getDataServices(): array
     {
--- a/src/DataService/DataServiceDiscovery.php
+++ b/src/DataService/DataServiceDiscovery.php
@@ -94,7 +94,7 @@ class DataServiceDiscovery
                 continue;
             }
 
-            $this->dataServices[$annotation->getService()][$annotation->getPath()][$annotation->getSchemaVersion()] = new DiscoveredDataService($class, $annotation);
+            $this->dataServices[$annotation->getService()][$annotation->getObjectType()][$annotation->getSchemaVersion()] = new DiscoveredDataService($class, $annotation);
         }
     }
 
--- a/src/DataService/DataServiceDiscovery.php
+++ b/src/DataService/DataServiceDiscovery.php
@@ -42,7 +42,7 @@ class DataServiceDiscovery
     /**
      * The array of discovered data services.
      *
-     * @var array
+     * @var DiscoveredDataService[]
      */
     private $dataServices = [];
 
@@ -64,8 +64,10 @@ class DataServiceDiscovery
 
     /**
      * Returns all the data services.
+     *
+     * @return DiscoveredDataService[] An array of all discovered data services.
      */
-    public function getDataServices()
+    public function getDataServices(): array
     {
         if (!$this->dataServices) {
             $this->discoverDataServices();
@@ -76,8 +78,6 @@ class DataServiceDiscovery
 
     /**
      * Discovers data services.
-     *
-     * @todo Return a structured class?
      */
     private function discoverDataServices()
     {
@@ -88,16 +88,13 @@ class DataServiceDiscovery
         /** @var SplFileInfo $file */
         foreach ($finder as $file) {
             $class = $this->classForFile($file);
+            /* @var \Lullabot\Mpx\DataService\Annotation\DataService $annotation */
             $annotation = $this->annotationReader->getClassAnnotation(new \ReflectionClass($class), 'Lullabot\Mpx\DataService\Annotation\DataService');
             if (!$annotation) {
                 continue;
             }
 
-            /* @var \Lullabot\Mpx\DataService\Annotation\DataService $annotation */
-            $this->dataServices[$annotation->getService()][$annotation->getPath()] = [
-                'class' => $class,
-                'annotation' => $annotation,
-            ];
+            $this->dataServices[$annotation->getService()][$annotation->getPath()][$annotation->getSchemaVersion()] = new DiscoveredDataService($class, $annotation);
         }
     }
 
--- a/src/DataService/DiscoveredDataService.php
+++ b/src/DataService/DiscoveredDataService.php
@@ -2,16 +2,21 @@
 
 namespace Lullabot\Mpx\DataService;
 
+use Lullabot\Mpx\DataService\Annotation\DataService;
+
 class DiscoveredDataService
 {
-
     /**
      * @var string
      */
     private $class;
+
+    /**
+     * @var DataService
+     */
     private $annotation;
 
-    public function __construct(string $class, $annotation)
+    public function __construct(string $class, DataService $annotation)
     {
         $this->class = $class;
         $this->annotation = $annotation;
@@ -26,7 +31,7 @@ class DiscoveredDataService
     }
 
     /**
-     * @return mixed
+     * @return DataService
      */
     public function getAnnotation()
     {
--- a/src/DataService/DiscoveredDataService.php
+++ b/src/DataService/DiscoveredDataService.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Lullabot\Mpx\DataService;
+
+class DiscoveredDataService
+{
+
+    /**
+     * @var string
+     */
+    private $class;
+    private $annotation;
+
+    public function __construct(string $class, $annotation)
+    {
+        $this->class = $class;
+        $this->annotation = $annotation;
+    }
+
+    /**
+     * @return string
+     */
+    public function getClass(): string
+    {
+        return $this->class;
+    }
+
+    /**
+     * @return mixed
+     */
+    public function getAnnotation()
+    {
+        return $this->annotation;
+    }
+}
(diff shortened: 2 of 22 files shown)


src/DataService/NotificationListener.php
```diff
@@ -95,16 +103,22 @@ public function sync()
     /**
      * Listen for notifications.
      *
+     * This method always filters to the object type defined in the discovered
+     * data service passed in the constructor, such as 'Media'.
```
deviantintegral, line 107: This is why I changed `path` to `objectType`.

src/DataService/NotificationListener.php
```diff
@@ -65,7 +73,7 @@ public function __construct(string $clientId, AuthenticatedClient $session, stri
         $this->resolveDomain = new ResolveDomain($this->authenticatedClient);
 
         /** @var ResolveAllUrls $resolver */
-        $resolver = ResolveAllUrls::load($this->authenticatedClient, $this->service)->wait();
+        $resolver = ResolveAllUrls::load($this->authenticatedClient, $this->service->getAnnotation()->getService(true))->wait();
```
deviantintegral, line 76: This ensures we always use the read-only endpoint, as documented by MPX.

src/DataService/NotificationListener.php
```diff
@@ -52,11 +56,15 @@ class NotificationListener
     /**
      * NotificationListener constructor.
      *
-     * @param string                            $clientId A string to identify this client in debugging.
+     * This method always uses the read-only version of a service.
+     *
+     * @see https://docs.theplatform.com/help/wsf-subscribing- to -change-notifications#tp-toc2
```
deviantintegral: Huh, I wonder what happened to this link. Will fix.

src/DataService/NotificationListener.php
```diff
@@ -95,16 +103,22 @@ public function sync()
...
      * @todo Add support for a configurable timeout?
-     * @todo Convert the return to an ObjectList.
      *
-     * @see \Lullabot\Mpx\DataService\NotificationListener::sync
+     * @see  \Lullabot\Mpx\DataService\NotificationListener::sync
+     * @see  \Lullabot\Mpx\DataService\Notification
      *
-     * @param int $since The last sync ID.
+     * @param int $since   The last sync ID.
+     * @param int $maximum (optional) The maximum number of notifications to return. Defaults to 500.
```
deviantintegral, line 117: 500 is also what MPX will limit to if you don't specify a size.

deviantintegral:
This is ready for review @juanolalla .

src/DataService/DataServiceDiscovery.php
```diff
@@ -64,8 +64,10 @@ public function __construct($namespace, $directory, $rootDir, Reader $annotation
 
     /**
      * Returns all the data services.
+     *
+     * @return DiscoveredDataService[] An array of all discovered data services.
```
juanolalla: We could document a little better how that array is expected to be structured. In line 97 I see it has three levels, the service, the object type and the schema version, it's not straightforward to see that.

src/DataService/DiscoveredDataService.php
```diff
@@ -0,0 +1,40 @@
...
+     * @var string
+     */
+    private $class;
+
+    /**
+     * @var DataService
+     */
+    private $annotation;
+
+    public function __construct(string $class, DataService $annotation)
```
juanolalla, line 19: Why are we calling annotation to the DataService? It looks confusing

src/DataService/NotificationListener.php
```diff
@@ -95,16 +103,22 @@ public function sync()
...
      *
-     * @return \GuzzleHttp\Promise\PromiseInterface A promise returning a list of notified objects.
+     * @return \GuzzleHttp\Promise\PromiseInterface A promise returning an array of Notifications.
```
juanolalla: Do you mean "returning an array of Notification objects"?

tests/src/JsonResponse.php
```diff
@@ -15,9 +15,9 @@ class JsonResponse extends Response
     public function __construct($status = 200, array $headers = [], $body = null, $version = '1.1', $reason = null)
     {
         if (isset($body)) {
-            if (is_file($body)) {
+            if (is_string($body) && is_file($body)) {
```
juanolalla, line 18: How could $body be string and file at the same time?

tests/src/Unit/DataService/NotificationListenerTest.php
```diff
@@ -0,0 +1,104 @@
...
+        /** @var StoreInterface|\PHPUnit_Framework_MockObject_MockObject $store */
+        $store = $this->getMockBuilder(StoreInterface::class)->getMock();
+        $session = new UserSession($user, $client, $store, $tokenCachePool);
+        $authenticatedClient = new AuthenticatedClient($client, $session);
+        $manager = DataServiceManager::basicDiscovery();
+        $service = $manager->getDataService('Media Data Service', 'Media', '1.10');
+        $listener = new NotificationListener($authenticatedClient, $service, 'unit-tests');
+        /** @var \Lullabot\Mpx\DataService\Notification[] $notifications */
+        $notifications = $listener->listen(98765)->wait();
+        $this->assertCount(2, $notifications);
```
juanolalla, line 82: Why this should be 2?

src/DataService/NotificationListener.php
juanolalla: Don't forget to fix it!

juanolalla:
That's a lot of awesome work! I've left a few comments and questions.

juanolalla (dismissed):
That's a lot of awesome work! I've left a few comments and questions.

src/DataService/NotificationListener.php
deviantintegral: Right, I meant that by capitalizing Notifications, but that's totally not obvious.

tests/src/JsonResponse.php
deviantintegral: It can't, but the method is suppose to allow you to pass one of:

1. A file name in the `fixtures` folder.
1. A raw JSON string.
1. An array of data to automatically json_encode.

The lack of an `is_string` check was breaking the third case as an array was being passed into `is_file`.

tests/src/Unit/DataService/NotificationListenerTest.php
deviantintegral: Unfortunately, the structure of notifications is slightly different from object lists. Where object lists have something like:

```json
{
  "entryCount": 2,
  "entries": [
    { "object..."  },
  ]
}
```

Notifications instead have an outer array only, with objects inside. See [notification.json](https://github.com/Lullabot/mpx-php/blob/e20144c91e705c87a8a88ec4d506cfe27b938c72/tests/fixtures/notification.json).

So in this case, we test MPX returning multiple notifications in one response (it can return up to the amount specified in the notification call).

src/DataService/DiscoveredDataService.php
deviantintegral: Oh wait, that's because it literally is an instance of the annotation class: `\Lullabot\Mpx\DataService\Annotation\DataService`

deviantintegral:
This is ready @juanolalla !

src/DataService/DiscoveredDataService.php
juanolalla: :+1:

tests/src/JsonResponse.php
juanolalla: But still, I don't see how this if could be ever be true, if is_file($body) were true as we expected, is_string($body) would be false, so it could never enter that condition...

tests/src/Unit/DataService/NotificationListenerTest.php
juanolalla: I didn't see the amount of 2 in the notification call. You mean listen() right? I don't see a second parameter there.

juanolalla:
Just that composed if where only one can be true and it would never reach the line within is driving me mad. Either is wrong or I'm confused and I can't see why.

tests/src/Unit/DataService/NotificationListenerTest.php
deviantintegral: Yes, "listen" eventually returns an array of Notifications. The individual HTTP response that's mocked contains two notifications (in production, you'd generally have anywhere from 1 to 500 items, or a timeout would trigger).

tests/src/JsonResponse.php
deviantintegral: Oh, the first if actually checks for any file, such as an absolute path. It's the second if that takes a raw file name and looks in the fixtures directory.

juanolalla: Ah ok! Sorry for that, so a file is actually a string and a file path, so is_file() returns true. I should have checked the is_file() documentation before asking this question.

juanolalla:
Sorry for my delay asking questions, now it have conflicts. Feel free to resolve and merge!

PR #70, opened 2018-03-28
Support insecure data services

### LINKED ISSUE #31: Implement the ResolveDomain method
It turns out the Drupal 7 module's method of hardcoding complete URLs for requests is not supported by MPX. In fact, the path to a service may change on an account-by-account basis.

We need to implement [ResolveDomain](https://docs.theplatform.com/help/wsf-resolvedomain-method) before we can finish #27.

deviantintegral:
Resolves #31.

This PR adds support for marking a service as "insecure", meaning it is only available over http and not https. As well, it uses Symfony's serializer for decoding the response, specifically so we can use the existing class to convert strings into Uri objects. This is particularly useful for converting service urls to HTTPS, which in general are available but are not discoverable via the API.

Also, if MPX updates their API so everything is available over HTTPS, we can simply ignore the 'insecure' marking in objects without breaking compatibility.

### DIFF
--- a/src/DataService/Annotation/DataService.php
+++ b/src/DataService/Annotation/DataService.php
@@ -71,28 +71,6 @@ class DataService
     }
 
     /**
-     * Return the relative path of the data service, such as '/data/Media'.
-     *
-     * @see https://docs.theplatform.com/help/wsf-how-to-find-the-url-of-a-service-in-the-service-registry
-     *
-     * @return string
-     */
-    public function getPath(): string
-    {
-        return $this->path;
-    }
-
-    /**
-     * Return the schema version this class implements, such as '1.10'.
-     *
-     * @return string
-     */
-    public function getSchemaVersion(): string
-    {
-        return $this->schemaVersion;
-    }
-
-    /**
      * @return string
      */
     public function getBaseUri(): string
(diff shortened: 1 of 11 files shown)


src/DataService/Annotation/DataService.php
```diff
@@ -97,4 +104,14 @@ public function hasBaseUri(): bool
     {
         return (bool) $this->baseUri;
     }
+
+    /**
+     * Return if this service is only available over HTTP.
+     *
+     * @return bool
+     */
+    public function isInsecure(): bool
```
juanolalla: We already have $insecure as a public property, why do you implement this?

src/DataService/ResolveDomainResponseExtractor.php
```diff
@@ -0,0 +1,34 @@
...
+use GuzzleHttp\Psr7\Uri;
+use Symfony\Component\PropertyInfo\PropertyTypeExtractorInterface;
+use Symfony\Component\PropertyInfo\Type;
+
+/**
+ * Extracts URIs with string array keys.
+ *
+ * Unfortunately, PhpDocExtractor's code uses private variables and final
+ * classes, so we can't override it to allow string array keys. Since this API
+ * only has one key, it is simpler to to just hardcode the property type.
```
juanolalla: The word "to" is repeated in this line

juanolalla (changes requested):

src/DataService/Annotation/DataService.php
deviantintegral: Total and complete habit from creating protected / private properties. Will remove!

deviantintegral:
Ready again @juanolalla .

PR #76, opened 2018-04-05
Fix null fatal when no range is set on ByFields

deviantintegral:
Ranges are optional, and mpx will automatically set a `1-500` range if one isn't requested. This PR fixes making queries that don't specify a range.

### DIFF
--- a/src/DataService/ByFields.php
+++ b/src/DataService/ByFields.php
@@ -34,10 +34,14 @@ class ByFields
      */
     protected $range;
 
+    /**
+     * ByFields constructor.
+     */
     public function __construct()
     {
         $this->sort = new Sort();
         $this->sort->addSort('id');
+        $this->range = new Range();
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/DataService/ByFields.php
```diff
@@ -34,10 +34,14 @@ class ByFields
...
     protected $range;
 
+    /**
+     * ByFields constructor.
+     */
     public function __construct()
     {
         $this->sort = new Sort();
         $this->sort->addSort('id');
+        $this->range = new Range();
```
deviantintegral, line 44: This fixes a null dereference on `toQueryParts` on `$this->range`.

src/DataService/Range.php
```diff
@@ -66,7 +66,7 @@ public function setEndIndex($endIndex): self
      *
      * @return Range A new range.
      */
-    public static function nextRange(ObjectList $list)
+    public static function nextRange(ObjectList $list): self
```
juampynr, line 69: Ah! Return types. Nice!

PR #88, opened 2018-04-06
Fix ad policy property being in the base data trait

deviantintegral:
The policy property was accidentally put into the BaseDataTrait, when it's totally not common to all objects. Also, the get / set methods were duplicated in the Media and Player classes, so I pulled all three out into a trait.


juampynr:
Looks good to me, @deviantintegral

PR #89, opened 2018-04-09
Support MediaFiles and Many More

deviantintegral:
Fixing nested objects in arrays meant adding support for way more objects than I initially thought I'd have to do in a single PR, but here it is! I will annotate with what files are generated and can probably be skimmed in review.


command/csv/account-object.csv
```diff
@@ -1,20 +1,20 @@
-FIELD NAME,ATTRIBUTES,DATA TYPE,DESCRIPTION
```
deviantintegral, line 1: I fixed the line endings, converting to unix-style in all of the CSV.

src/DataService/CachingContextFactory.php
```diff
@@ -0,0 +1,203 @@
...
+     * @param string $namespace    It does not matter if a `\` precedes the namespace name, this method first normalizes.
+     * @param string $fileContents the file's contents to retrieve the aliases from with the given namespace.
+     *
+     * @see Context for more information on Contexts.
+     *
+     * @return Context
+     */
+    public function createForNamespace($namespace, $fileContents)
+    {
+        static $cache = [];
```
deviantintegral, line 57: This is the reason I had to copy this class. The details?

I had to switch from using the ReflectionExtractor to the PhpDocExtractor. That means that Symfony discovers what types of objects a property accepts from the phpdocs instead of the phpcode. Unfortunately, PHP itself doesn't have generics (the C# / Java term) meaning "this property accepts an array of Ratings". PHP code can only indicate that it at accepts an array, but not what type each array item must be.

The PHPDoc extractor lets us write something like `\DateTime[]` and Symfony will know that all set calls for that property must be an array of \DateTime objects.

If you read this function, you'll see it re-parses `$fileContents` every time it's called. So, for a MediaFile with tons of properties, it's hundreds of times which is totally not OK for a first-time load. Since this method knows that `$fileContents` will always parse identically, we can do a static cache off of it's value. Before this cache, it was a huge performance hit here - 45-90 seconds!

Why a basic static variable, and not a proper cache backend? We are already wrapping this in a cache backend, but that cache set only happens after a class has been parsed completely. It doesn't incrementally set the cache as it goes.

src/DataService/CachingPhpDocExtractor.php
```diff
@@ -0,0 +1,245 @@
...
+     * @param string[]|null            $arrayMutatorPrefixes
+     */
+    public function __construct(DocBlockFactoryInterface $docBlockFactory = null, array $mutatorPrefixes = null, array $accessorPrefixes = null, array $arrayMutatorPrefixes = null)
+    {
+        if (!class_exists(DocBlockFactory::class)) {
+            throw new \RuntimeException(sprintf('Unable to use the "%s" class as the "phpdocumentor/reflection-docblock" package is not installed.', __CLASS__));
+        }
+
+        $this->docBlockFactory = $docBlockFactory ?: DocBlockFactory::createInstance();
+        $this->contextFactory = new CachingContextFactory();
```
deviantintegral, line 50: This is why I had to copy this class. The upstream class declares this property as a private variable, which means we can't change it just by overriding the constructor. If only it was protected!

src/DataService/CachingPhpDocExtractor.php
```diff
@@ -0,0 +1,245 @@
...
+    private function getDocBlockFromProperty($class, $property)
+    {
+        // Use a ReflectionProperty instead of $class to get the parent class if applicable
+        try {
+            $reflectionProperty = new \ReflectionProperty($class, $property);
+        } catch (\ReflectionException $e) {
+            return;
+        }
+
+        return $this->docBlockFactory->create($reflectionProperty, $this->contextFactory->createFromReflector($reflectionProperty->getDeclaringClass()));
```
deviantintegral, line 202: And, the one use of `$this->contextFactory` is _also_ in a private method, so once we would have to override all of the public methods we may as well just copy the whole class

src/DataService/Media/AvailabilityWindow.php
```diff
@@ -0,0 +1,90 @@
+<?php
```
deviantintegral, line 1: All of these next classes are generated code. It would be OK to skip them and go to just the test classes below.

tests/src/Functional/DataService/Media/MediaFileTest.php
```diff
@@ -0,0 +1,39 @@
...
+        $filter = new ByFields();
+        $range = new Range();
+        $range->setStartIndex(1)
+            ->setEndIndex(2);
+        $filter->setRange($range);
+        $results = $dof->select($filter, $this->account);
+
+        foreach ($results as $index => $result) {
+            $this->assertInstanceOf(UriInterface::class, $result->getId());
+            if ($index > 2) {
```
deviantintegral, line 34: This makes me think I should consider implementing \ArrayAccess on $results too. However, it could cause memory problems as to rewind the array you'd have to keep all the results in memory, and not just those from the current page. Thoughts?

tests/src/Unit/DataService/CachingContextFactoryTest.php
```diff
@@ -0,0 +1,205 @@
+<?php
```
deviantintegral, line 1: This was fun. I basically copied the test class from phpDoc, and proved I didn't break anything with the caching as the test didn't require any changes.

src/DataService/BaseDataTrait.php
```diff
@@ -2,8 +2,6 @@
 
 namespace Lullabot\Mpx\DataService;
 
-use Psr\Http\Message\UriInterface;
```
juanolalla, line 5: Why aren't you doing here the same you did in src/DataService/AdPolicyDataTrait.php? I mean leaving that use clause and shortening all the rest of @var ... uses.

src/DataService/CachingContextFactory.php
```diff
@@ -0,0 +1,203 @@
...
+
+    /** The literal used between sets of use statements */
+    const T_LITERAL_USE_SEPARATOR = ',';
+
+    /**
+     * Build a Context given a Class Reflection.
+     *
+     * @param \Reflector $reflector
+     *
+     * @see Context for more information on Contexts.
```
juanolalla, line 25: This sentence seems redundant/confusing, it could be clearer like this:
`@see phpDocumentor\Reflection\Types\Context for more information on contexts.`

src/DataService/CachingContextFactory.php
```diff
@@ -0,0 +1,203 @@
...
+final class CachingContextFactory
+{
+    /** The literal used at the end of a use statement. */
+    const T_LITERAL_END_OF_USE = ';';
+
+    /** The literal used between sets of use statements */
+    const T_LITERAL_USE_SEPARATOR = ',';
+
+    /**
+     * Build a Context given a Class Reflection.
```
juanolalla, line 21: I think you meant a class "Reflector"

src/DataService/CachingContextFactory.php
```diff
@@ -0,0 +1,203 @@
...
+     *
+     * @return Context
+     */
+    public function createFromReflector(\Reflector $reflector)
+    {
+        if (method_exists($reflector, 'getDeclaringClass')) {
+            $reflector = $reflector->getDeclaringClass();
+        }
+
+        $fileName = $reflector->getFileName();
```
juanolalla, line 35: Why do you check if the class implementing \Reflector has the getDeclaringClass() method while you don't check here and in the next line if the class has these other two methods?

src/DataService/CachingContextFactory.php
```diff
@@ -0,0 +1,203 @@
...
+        return new Context($namespace, []);
+    }
+
+    /**
+     * Build a Context for a namespace in the provided file contents.
+     *
+     * @param string $namespace    It does not matter if a `\` precedes the namespace name, this method first normalizes.
+     * @param string $fileContents the file's contents to retrieve the aliases from with the given namespace.
+     *
+     * @see Context for more information on Contexts.
```
juanolalla, line 51: Same here, maybe this @see worth the whole namespace for better clarity, and the word contexts in lower case. What do you think?

src/DataService/CachingContextFactory.php
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        return new Context($namespace, []);
+    }
+
+    /**
+     * Build a Context for a namespace in the provided file contents.
+     *
+     * @param string $namespace    It does not matter if a `\` precedes the namespace name, this method first normalizes.
+     * @param string $fileContents the file's contents to retrieve the aliases from with the given namespace.
```
juanolalla, line 49: The sentence should start with a capital "T". Also, I think that "from with" is not what you wanted? Remove "with" perhaps?

src/DataService/CachingContextFactory.php
```diff
@@ -0,0 +1,203 @@
...
+        $key = md5($fileContents);
+
+        if (!isset($cache[$namespace][$key])) {
+            $namespace = trim($namespace, '\\');
+            $useStatements = [];
+            $currentNamespace = '';
+            $tokens = new \ArrayIterator(token_get_all($fileContents));
+
+            while ($tokens->valid()) {
+                switch ($tokens->current()[0]) {
```
juanolalla, line 67: According to `token_get_all()` documentation:

>  An array of token identifiers. Each individual token identifier is either a single character (i.e.: ;, ., >, !, etc...), or a three element array containing the token index in element 0, the string content of the original token in element 1 and the line number in element 2. 

So I wonder if in the case of a single character, that [0] could throw an error.

tests/src/Functional/DataService/Media/MediaFileTest.php
juanolalla: I'm not familiar with the use of ArrayAccess interface, not sure what pros and cons it could have.

juanolalla (commented):
Wow! This was huge, awesome work. I couldn't make a deep review since there are many things and it would take hours to understand everything, but I did what I could! Some minor things and some questions. Feel free to merge if you need it since I won't be working tomorrow Friday.

tests/src/Functional/DataService/Media/MediaFileTest.php
juampynr: I suggest that you create an issue to work on it at a later stage.

juampynr:
83% coverage! Well done :+1:

juampynr:
@deviantintegral this pr looks good to me. Feel free to merge.

tests/src/Functional/DataService/Media/MediaFileTest.php
deviantintegral: Filed #90 !

src/DataService/BaseDataTrait.php
deviantintegral: I'm totally not sure. Probably a mistake on my part. Fixing!

src/DataService/CachingContextFactory.php
deviantintegral: Good catch!

deviantintegral: Actually, this is verbatim in the original class so I'd like to keep it the same.

src/DataService/CachingContextFactory.php
deviantintegral: This is directly copied from https://github.com/phpDocumentor/TypeResolver/blob/master/src/Types/ContextFactory.php so while it's a little rough to read I'd rather it be the same so future diffs are easier to follow against the upstream class, should we ever need to update it.

src/DataService/CachingContextFactory.php
deviantintegral: Sorry, I should have made it clear that the only method I modified is `\Lullabot\Mpx\DataService\CachingContextFactory::createForNamespace`. Do you want to file your questions upstream at https://github.com/phpDocumentor/TypeResolver ?

src/DataService/CachingContextFactory.php
deviantintegral: Oooo nice find. Do you think you could come up with a failing test case in the upstream project?

src/DataService/CachingContextFactory.php
deviantintegral: Seems worthy of an upstream PR?

src/DataService/BaseDataTrait.php
deviantintegral: I forgot that the phpdocumentor reflection was breaking, thinking that `UriInterface` was then part of the same namespace as the trait. I _think_ this is an upstream bug, and I will file an issue on our side to investigate.

src/DataService/CachingContextFactory.php
juanolalla: :+1:

src/DataService/CachingContextFactory.php
juanolalla: :+1:

src/DataService/CachingContextFactory.php
juanolalla: No, it's ok, we better keep it like the original, I didn't realize that part wasn't your code

src/DataService/CachingContextFactory.php
juanolalla: Maybe, I'll keep that in mind for this week

PR #98, opened 2018-04-23
Support loading enabled Player plugins

deviantintegral:


juampynr (approved):
Static review looks fine. Tests pass and coverage is still high. Nice job @deviantintegral!

PR #99, opened 2018-04-23
Add an ObjectInterface for all mpx objects

deviantintegral:
Now that I'm working on importing Player data, we have a need to share code between Media and Player objects in the downstream Drupal module. This PR simplifies code there by allowing it to typehint to an `ObjectInterface` to get the base data for any mpx object.

This PR:

* Removes the `CreateKeyInterface` that it turns out we don't need. If we need to document compound keys, we will use the annotations instead. If we need to add support for custom keys, those will become their own interface.
* Converts `BaseDataTrait` to an abstract class. This is primarily so it can implement `ObjectInterface`.


src/DataService/ObjectInterface.php
```diff
@@ -0,0 +1,65 @@
...
+
+namespace Lullabot\Mpx\DataService;
+
+/**
+ * Defines an interface object properties common to all mpx objects.
+ */
+interface ObjectInterface
+{
+    /**
+     * Returns The date and time that this object was created.
```
juanolalla: "Returns the" instead of "Returns The". This is happening too in the following comments of properties

juanolalla (changes requested):
Just the comments fixes, for the rest I see it all right :+1:

deviantintegral:
Good catch. Fixed and merged!

PR #100, opened 2018-04-23
Always use https

deviantintegral:
~~Depends on #99.~~

After discussing with thePlatform, they say that all services should be available over https, even if their docs are not updated reflecting that. This PR:

* Removes the `insecure` annotation for a data service, which we previously used for the Player data r/w service, as that service is now available over https.
* When loading mpx objects, we always convert URLs to https. I thought about doing it at the client layer, but I think as a developer if you use the client directly you expect the protocol of your request to remain untouched.

### DIFF
--- a/src/DataService/Annotation/DataService.php
+++ b/src/DataService/Annotation/DataService.php
@@ -46,13 +46,6 @@ class DataService
     public $baseUri;
 
     /**
-     * Is this service only available over HTTP?
-     *
-     * @var bool
-     */
-    public $insecure = false;
-
-    /**
      * Return the service of the data object, such as 'Media Data Service'.
      *
      * @param bool $readonly (optional) Set to true to return the read-only name of the service.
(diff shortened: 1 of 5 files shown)


src/DataService/DataServiceManager.php
```diff
@@ -68,7 +68,7 @@ public function getDataService(string $name, string $objectType, string $schema)
      *
      * @throws \RuntimeException
      *
-     * @return object
+     * @return ObjectInterface
```
juampynr: :+1:

src/DataService/Media/Chapter.php
```diff
@@ -87,7 +87,7 @@ public function getThumbnailUrl(): Uri
     /**
      * Set a link to the thumbnail image for this chapter.
      *
-     * @param Uri
+     * @param Uri $thumbnailUrl The URL of the thumbnail.
```
juampynr: Should the description go in the next line?

deviantintegral: I'm following the Symfony style guides for the library. See [.php_cs.dist](https://github.com/Lullabot/mpx-php/blob/master/.php_cs.dist).

juampynr: Gotcha

deviantintegral:
@juampynr Ready again!

juampynr:
Looks good :+1:

PR #101, opened 2018-04-26
Support generating player URLs for iframes

deviantintegral:
The `PublicIdentifierInterface` has been added so the Drupal module can implement it on the Account config entity object, allowing it to be directly passed in to the PHP library.

I also fixed the Account object duplicating some code that is now available in the `ObjectBase` class.

### DIFF
--- a/src/DataService/PublicIdentifierInterface.php
+++ b/src/DataService/PublicIdentifierInterface.php
@@ -13,5 +13,4 @@ interface PublicIdentifierInterface
      * @return string
      */
     public function getPid(): string;
-
 }
--- a/src/DataService/PublicIdentifierInterface.php
+++ b/src/DataService/PublicIdentifierInterface.php
@@ -13,4 +13,11 @@ interface PublicIdentifierInterface
      * @return string
      */
     public function getPid(): string;
+
+    /**
+     * Set the public identifier for this mpx object.
+     *
+     * @param string
+     */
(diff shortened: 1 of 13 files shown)


src/DataService/PublicIdentifierInterface.php
```diff
@@ -0,0 +1,16 @@
...
+ */
+interface PublicIdentifierInterface
+{
+    /**
+     * Returns the public identifier for this mpx object.
+     *
+     * @return string
+     */
+    public function getPid(): string;
+}
```
juanolalla: Wouldn't it be appropriate to include setPid() in this interface too?

juanolalla (commented):
Awesome work! Just a question/suggestion, but feel free to merge

src/DataService/PublicIdentifierInterface.php
deviantintegral: I'd actually removed it over at https://github.com/Lullabot/mpx-php/pull/101/commits/19d5af2f68c33af74fa5bd96e545ab4217f3bbd4

However, I realized that while splitting the interface makes things a little simpler for read-only clients, it makes them more complex for clients that want to modify data. I'll put it back!

deviantintegral:
Thanks Juan! I added the interface method back, and I added a trait too to reduce some code duplication.

PR #102, opened 2018-04-26
Fix account incorrectly using the user ID

### LINKED ISSUE #93: Fix AuthenticatedClient::getAccount
`\Lullabot\Mpx\AuthenticatedClient::getAccount()` is completely wrong - it takes a _User_ ID and set it as an _Account_ ID. I don't think this ever worked, and I think this implies that our helpful "optional account context" in `DataObjectFactory::select()` is going to have to be removed.

deviantintegral:
This fixes #93 .

While an account context is required for almost all data service requests, it is not required for Account requests (since you may not know the account to use until you fetch accounts). There was a mixup in the `getAccount` method where it actually used a User ID, and not an Account ID.

This PR removes that, and makes account contexts optional, with a bit of documentation that they are almost always required.

### DIFF
--- a/src/AuthenticatedClient.php
+++ b/src/AuthenticatedClient.php
@@ -5,7 +5,6 @@ namespace Lullabot\Mpx;
 use GuzzleHttp\ClientInterface;
 use GuzzleHttp\Promise\Promise;
 use GuzzleHttp\Promise\PromiseInterface;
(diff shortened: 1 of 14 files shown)


juampynr:
Updated branch with master. Waiting for test results.

Code looks fine.

PR #103, opened 2018-04-29
Limit queries to 100 items by default

deviantintegral:
As I was testing importing media objects with the Drupal library, I kept seeing memory leaks. Each result page that was loaded would take an additional 30-40MB of memory, across multiple PHP versions and operating systems. Both xhprof and Blackfire showed the memory use as being within PHP's own C code in `json_decode()`. Thinking it was still in our code somehow, I reduced the result set to try and create more objects and run loops more often, hoping to make the leak worse and easier to fix.

However, I found that once I got below a certain amount, PHP's memory use would stabilize instead of growing. For example, at 100 items per page, PHP would only use ~60MB of memory, and happily stay there.

I need to dig in more, probably starting with [the Zend Memory Manager](http://www.phpinternalsbook.com/php7/memory_management/zend_memory_manager.html), but I figured I'd get a PR up that's good enough for our purposes.

Some quick testing doesn't show any real performance hit by using smaller result sets (and makes things less bursty too).

### DIFF
--- a/src/DataService/ByFields.php
+++ b/src/DataService/ByFields.php
@@ -9,6 +9,10 @@ namespace Lullabot\Mpx\DataService;
  * specified. Since sorting may be inconsistent across pages, this class will
  * automatically sort by 'id' if no sort is specified.
  *
+ * By default, pages are set to 100 items per page. This matches well with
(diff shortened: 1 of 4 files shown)


juampynr (commented):
Looks good. Merging.

PR #106, opened 2018-05-03
Support caching of resolveDomain responses

deviantintegral:
This saves around 500ms for every single mpx request, once the cache has been populated. Once this in, we will be able to wire up Drupal's cache in the Drupal module just like we do with the token cache.


juanolalla (approved):
Good job!!

PR #107, opened 2018-05-03
Add a method to yield pages in lists

deviantintegral:
We would like to support concurrent requests as there's some clear performance benefits in importing mpx data.

Initially, I thought we could build an array of promises to `ObjectLists`, and pass them to `each_limit`:

```
<?php
    each_limit($promises_to_lists, 10, function(ObjectList $list) use ($media_type, $importer) {
      foreach ($list as $item) {
        ...
      }
    })->wait();
```

However, this was causing my initial tests to hang. After debugging, I discovered that Guzzle's curl handler doesn't actually respect the limit passed in to each_limit. Instead, all (in this case, 700+) requests are opened, and they are all attempted in parallel. That's because each call to generate a Request object automatically adds it to the list of curl handles to be executed.

In other words, for the first time I found a real-world case to use generators and the `yield` keyword.

With `yield`, each_limit ensures that there are only 10 outstanding promises total, and additional promises are generated as needed. This also helps a bit with memory use too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
+- Add a method to yield pages in lists #107
+
 ### Changed
 
 ### Removed
--- a/src/DataService/DataObjectFactory.php
+++ b/src/DataService/DataObjectFactory.php
@@ -186,7 +186,7 @@ class DataObjectFactory
             ],
         ];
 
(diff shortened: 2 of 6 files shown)


juampynr:
> After debugging, I discovered that Guzzle's curl handler doesn't actually respect the limit passed in to each_limit. Instead, all (in this case, 700+) requests are opened, and they are all attempted in parallel.

That must have been tricky to debug.

juampynr:
Looks good and tests pass, plus coverage is still high. It was nice learning about `yield`!

PR #108, opened 2018-05-08
Fix nextlist returning true for single-item lists

deviantintegral:
https://github.com/Lullabot/mpx-php/pull/108/commits/c5a65033971e0c76f2e480dba17b17bb0532afec is the test case that currently fails on master, and is fixed by this PR.


juanolalla (approved):
Good work, thanks!

PR #113, opened 2018-05-11
Fix broken range calculations for the end of a list of ranges

deviantintegral:
I discovered this when loading some real content from mpx. The test at `testPartialEndRange()` currently fails on master, where too many ranges are returned and their indexes are beyond the end of the list. I also added a functional test showing what I was doing that triggered the bug (though it's dependent on the data in the mpx account you're testing with).


juampynr:
So many ranges. This made me think of Chuck Norris.

PR #114, opened 2018-05-11
Support caching of ResolveAllUrls API calls

### LINKED ISSUE #44: Add caching support for service resolution
There's no documentation stating what caching rules we can use, and the resolveDomain() and similar calls all return `no-cache`. However, we should be able to use a static cache at least, and if we follow the same pattern as the token cache it will be easy to swap out later.

deviantintegral:
This PR resolves #44.

* Enables cross-request caching of services when loading objects. Previously, we only actually enabled a basic array cache.
* Converts ResolveAllUrls to follow the same structure as ResolveDomain.
* Supports caching of ResolveAllUrls.

### DIFF
--- a/src/DataService/DataObjectFactory.php
+++ b/src/DataService/DataObjectFactory.php
@@ -60,22 +60,21 @@ class DataObjectFactory
     /**
      * DataObjectFactory constructor.
      *
-     * @todo Inject the resolveDomain() instead of constructing?
-     *
      * @param DiscoveredDataService             $dataService         The service to load data from.
      * @param \Lullabot\Mpx\AuthenticatedClient $authenticatedClient A client to make authenticated MPX calls.
-     * @param CacheItemPoolInterface|null       $cacheItemPool       (optional) Cache to store reflection metadata.
+     * @param CacheItemPoolInterface|null       $cacheItemPool       (optional) Cache to store API metadata.
      */
     public function __construct(DiscoveredDataService $dataService, AuthenticatedClient $authenticatedClient, CacheItemPoolInterface $cacheItemPool = null)
     {
         $this->authenticatedClient = $authenticatedClient;
-        $this->resolveDomain = new ResolveDomain($this->authenticatedClient);
         $this->dataService = $dataService;
 
         if (!$cacheItemPool) {
             $cacheItemPool = new ArrayCachePool();
         }
         $this->cacheItemPool = $cacheItemPool;
+
+        $this->resolveDomain = new ResolveDomain($this->authenticatedClient, $this->cacheItemPool);
     }
 
     /**
--- a/src/DataService/DataObjectFactory.php
+++ b/src/DataService/DataObjectFactory.php
@@ -220,6 +220,7 @@ class DataObjectFactory
             // If no account is specified, we must use the ResolveAllUrls service instead.
             if (!$account) {
                 $resolver = new ResolveAllUrls($this->authenticatedClient, $this->cacheItemPool);
+
                 return $resolver->resolve($annotation->getService($readonly))->resolve().$annotation->getPath();
             }
 
--- a/src/DataService/DataObjectFactory.php
+++ b/src/DataService/DataObjectFactory.php
@@ -221,7 +221,7 @@ class DataObjectFactory
             if (!$account) {
                 $resolver = new ResolveAllUrls($this->authenticatedClient, $this->cacheItemPool);
 
-                return $resolver->resolve($annotation->getService($readonly))->resolve().$annotation->getPath();
+                return $resolver->resolve($annotation->getService($readonly))->getUrl().$annotation->getPath();
             }
 
             $resolved = $this->resolveDomain->resolve($account);
--- a/src/DataService/DataObjectFactory.php
+++ b/src/DataService/DataObjectFactory.php
@@ -206,8 +206,6 @@ class DataObjectFactory
     /**
      * Get the base URI from an annotation or service registry.
      *
-     * @todo This should cache resolved URLs.
-     *
      * @param DataService $annotation The annotation data is being loaded for.
      * @param IdInterface $account    (optional) The account to use for service resolution.
      * @param bool        $readonly   (optional) Load from the read-only service.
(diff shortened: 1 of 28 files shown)


juampynr:
@deviantintegral can you fix conflicts in this branch?

deviantintegral:
Done! Feel free to resolve conflicts yourself too if they are simple, like in the CHANGELOG or similar.

src/Service/AccessManagement/ResolveAllUrlsResponse.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace Lullabot\Mpx\Service\AccessManagement;
+
+use Psr\Http\Message\UriInterface;
+
+class ResolveAllUrlsResponse
```
juanolalla: It would be good if we explained a little bit what is this class for, and how it is related to ResolveAllUrls

src/Service/AccessManagement/ResolveAllUrlsResponse.php
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * @param string $service
+     */
+    public function setService(string $service)
+    {
+        $this->service = $service;
+    }
+
+    /**
+     * @param bool $insecure (optional) Set to true to request the insecure version of this service.
```
juanolalla: I'm missing here a description of this method. It's probably clear to you, but since we're using ResolveAllUrls->resolve and ResolveAllUrls->resolve (and ResolveBase, ResolveDomain...) there are so many uses of the word resolve that isn't super clear to me what is the real use in each case.

src/DataService/DataObjectFactory.php
```diff
@@ -222,10 +219,9 @@ private function getBaseUri(DataService $annotation, IdInterface $account = null
         if (!($base = $annotation->getBaseUri())) {
             // If no account is specified, we must use the ResolveAllUrls service instead.
             if (!$account) {
-                /** @var ResolveAllUrls $urls */
-                $urls = ResolveAllUrls::load($this->authenticatedClient, $annotation->getService($readonly))->wait();
+                $resolver = new ResolveAllUrls($this->authenticatedClient, $this->cacheItemPool);
 
-                return $urls->resolve().$annotation->getPath();
+                return $resolver->resolve($annotation->getService($readonly))->resolve().$annotation->getPath();
```
juanolalla: This line looks a little bit confusing. Semantically, it's strange that we use the resolve() method twice, it's like we are resolving it first, but not completely, and then we do it again, over a different returned object.

To me it's also confusing that we are using the word Resolve also in the classes, when it's a verb. The fact that you're using here and in the tests the variable $resolver, it's a clue that we should probably use something like that in the class name. Like I mentioned above, it's also strange that ResolveAllUrls::resolve() returns a ResolveAllUrlsResponse, that also have a resolve() method! And the ResolveAllUrlsResponse doesn't inherit ResolveBase, which can be confusing too. The fact that you're using $resolved, and $response as variable names to store that object, makes me thing that we should probably change the class name accordingly.

Oh! And it's curious that ResolveBase doesn't provide a resolve() method. But yeah, it's not an interface and there is no shared logic in the resolve for Domain and AllUrls right?

In summary, it looks to me that we're abusing the word resolve in methods and classes, and that we are using class names that sounds like verbs/actions, which is more appropriate for method names.

As a suggested solution, we could better use ResolverBase, DomainResolver, AllUrlsResolver, for example, or even a clearer set of words, if we find it. And I think we should change the ResolveAllUrlsResponse (which could be AllUrlsResolverResponse?) resolve method to a different, more specific, set of words. On the other hand, if resolve() has better meaning in the response class, maybe we could change and extend with other words the use of resolve in the other classes. In any case, I think we should avoid doing resolve()->resolve().

juanolalla (approved):
Great work here Andrew!

I've left you some comments regarding clarification and renaming because I consider it confusing. In case you're agree to improve that naming, feel free to merge and open a new issue or pull request for that, or well make the changes in this same pull request before merging.

src/Service/AccessManagement/ResolveAllUrlsResponse.php
deviantintegral: Done!

src/Service/AccessManagement/ResolveAllUrlsResponse.php
deviantintegral: Good point. I've renamed resolve() here to getUrl().

src/Service/AccessManagement/ResolveAllUrlsResponse.php
juanolalla: :+1:

src/Service/AccessManagement/ResolveAllUrlsResponse.php
juanolalla: :+1:

juanolalla (approved):

src/DataService/DataObjectFactory.php
deviantintegral: > Semantically, it's strange that we use the resolve() method twice

Indeed. Does the rename to `getUrl` make this clearer?

> To me it's also confusing that we are using the word Resolve also in the classes, when it's a verb.

> As a suggested solution, we could better use ResolverBase, DomainResolver, AllUrlsResolver

My thinking was to match [the upstream mpx method names](https://docs.theplatform.com/help/wsf-resolveallurls-method). I agree, it would be better if they were REST-ish instead of RPC-ish in their naming, but I think it's better to match so anyone doing a class search based off of the mpx docs will find them.

> And the ResolveAllUrlsResponse doesn't inherit ResolveBase, which can be confusing too. 

Do think this it makes sense to move the response classes into their own directory / namespace?

> The fact that you're using $resolved, and $response as variable names to store that object, makes me thing that we should probably change the class name accordingly.

Oh, that's actually a bug in `ResolveDomain`. I mean, nothing is broken, but I didn't mean to overwrite the initial Guzzle response.

> Oh! And it's curious that ResolveBase doesn't provide a resolve() method. But yeah, it's not an interface and there is no shared logic in the resolve for Domain and AllUrls right?

Yeah, I thought about that, and the difference is that `resolveDomain` returns all options in one response, so we want to pass in the individual service we care about.

juanolalla: You're right, matching should be a priority, otherwise it would be more confusing. The getUrl() renaming does make it clearer :+1:

PR #117, opened 2018-05-11
Rename user to userSession

### LINKED ISSUE #75: Rename UserSession parameter in AuthenticatedClient
This should be `session` or `userSession`, not `user`:

https://github.com/Lullabot/mpx-php/blob/dfa0cd35856c9e82699bf4e76039954fd7d49d3a/src/AuthenticatedClient.php#L37

deviantintegral:
Resolves #75 .


juanolalla (approved):
Looks good!

PR #118, opened 2018-05-14
Fix using the number of entries in the current page instead of the total

deviantintegral:
`\Lullabot\Mpx\DataService\ObjectList::hasNext()` was mistakenly using the number of entries in the current page to determine if a next page existed. It should have been using the total number of results across all pages. Prior to #108 this worked as we checked if the entry count was less than the items specified per page, but left us open to the bug we fixed there.

I've added docs to clarify that entry count is only for the current page.

### DIFF
--- a/tests/src/Unit/DataService/ObjectListTest.php
+++ b/tests/src/Unit/DataService/ObjectListTest.php
@@ -79,14 +79,15 @@ class ObjectListTest extends TestCase
     {
         $this->list->setEntries([new \stdClass()]);
         $this->list->setStartIndex(1);
-        $this->list->setEntryCount(11);
+        $this->list->setEntryCount(10);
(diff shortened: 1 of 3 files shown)


tests/src/Unit/DataService/ObjectListTest.php
```diff
@@ -79,14 +79,15 @@ public function testHasNext()
     {
         $this->list->setEntries([new \stdClass()]);
         $this->list->setStartIndex(1);
-        $this->list->setEntryCount(11);
+        $this->list->setEntryCount(10);
```
deviantintegral, line 82: These tests by themselves fail on master, proving the bug.

juanolalla (approved):
:+1:

Issue #92, opened 2018-04-16
Support custom fields on mpx objects


deviantintegral:
mpx supports custom fields on objects, and includes an API for clients to determine what fields exist and what type of data each field is:

https://docs.theplatform.com/help/wsf-working-with-custom-fields

Custom fields reside in their own namespace, and multiple namespaces of custom fields may be available on a single object.

My first thought is to support this through a custom property info class, that calls the mpx API to find out all of the available fields. However, that hurts discoverability of the data in IDEs, as there won't be any sort of autocomplete since the property info is discovered at runtime.

Another option would be to allow other libraries to extend from our classes, using the annotations we already have to select a different class. However, that will still break IDE autocomplete without lots of `@var ...` comments.

A third option would be to somehow allow declaring custom fields by their namespace, but again I have no idea how to do that preserving autocomplete.

The "worst" option would be to just dump custom fields into an array of raw data, and allow library consumers to turn those into objects if they wish.

PR #119, opened 2018-05-14
Support generating custom field classes

deviantintegral:
This PR _mostly_ handles #92 by:

1. Adds support for loading custom field data from mpx. This ensures developers are working with custom fields as they are actually set configured.
1. Adds a console command to generate classes for custom fields. Each class represents a namespace.

```
bin/console help mpx:create-custom-field
Usage:
  mpx:create-custom-field <namespace> <data-service> <data-object> <schema>
  mpx:create-custom-field 'Lullabot\Mpx\CustomField' 'Media Data Service' 'Media' '1.10'

Arguments:
  namespace             The fully-qualified class name to generate. Do not include the leading slash, and wrap the class name in single-quotes to handle shell escaping.
  data-service          The data service to generate the class for, such as 'Media Data Service'.
  data-object           The object to generate the class for, such as 'Media'
  schema                The schema version of the object, such as '1.10'

Options:
  -h, --help            Display this help message
  -q, --quiet           Do not output any message
  -V, --version         Display this application version
      --ansi            Force ANSI output
      --no-ansi         Disable ANSI output
  -n, --no-interaction  Do not ask any interactive question
  -v|vv|vvv, --verbose  Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug

Help:
  This command generates a PHP class for a custom field implementation.

  A mpx username and password is required. They will be requested interactively,
  or MPX_USERNAME and MPX_PASSWORD environment variables will be used.

  As mpx provides no way to infer a reasonable class name, each class is named
  sequentially with the expectation that it will be renamed later. One file will
  be generated for each namespace of custom fields that is detected. Files will
  be written to the current working directory.
```

Still to be done is to add an annotation for custom field classes. That way, site-specific code can annotate a generated class as belonging to a specific namespace, and we can discover them automatically when deserializing a mpx response. I will do that as a new PR if this gets merged first.

### DIFF
--- a/command/csv/field-object.csv
+++ b/command/csv/field-object.csv
@@ -1,0 +1,27 @@
+FIELD NAME,ATTRIBUTES,DATA TYPE,DESCRIPTION
+added,,DateTime,"The date and time that this object was created
+ This field is queryable on the following endpoints only:AssetType/Field Category/Field Media/Field MediaFile/Field Release/Field Server/Field"
+addedByUserId,,URI,The id of the user that created this object
+allowedValues,,dataType[],The allowed values for this custom field
+dataStructure,,DataStructure,The data structure of this custom field
+dataType,,DataType,The data type of this custom field
+defaultValue,,dataType,The default value for this custom field
+description,,String,The description of this object
+fieldName,,String,The node name for this custom field in XML and JSON
+guid,,String,An alternate identifier for this object that is unique within the owning account
+id,,URI,The globally unique URI of this object
+includeInTextSearch,,Boolean,Whether this custom field is indexed for search
+isUnique,,Boolean,Whether this custom field stores unique values and functions as a create key
+length,,Integer,The maximum string length or the number of decimal positions allowed in this custom field's value
+locked,,Boolean,Whether this object currently allows updates
+namespace,,String,The namespace this custom field belongs to
+namespacePrefix,,String,An XML namespace prefix for this field
+notifyAlways,,Boolean,Whether this custom field is always available in change notifications
+notifyChanges,,Boolean,Whether this custom field is available on change in update notifications
+notifyDelete,,Boolean,Whether this custom field is available in delete notifications
+ownerId,,URI,The id of the account that owns this object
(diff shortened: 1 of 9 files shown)


command/csv/field-object.csv
```diff
@@ -0,0 +1,27 @@
+FIELD NAME,ATTRIBUTES,DATA TYPE,DESCRIPTION
```
juampynr, line 1: Is this a reference of fields, @deviantintegral ?

command/src/CreateCustomFieldClassCommand.php
```diff
@@ -0,0 +1,181 @@
...
+        'String' => 'string',
+        'Time' => '\\' . \DateTime::class,
+        'URI' => '\\' . UriInterface::class,
+    ];
+
+    protected function configure() {
+        $help = <<<EOD
+This command generates a PHP class for a custom field implementation.
+
+A mpx username and password is required. They will be requested interactively,
```
juampynr: Shouldn't this be `An`?

command/src/CreateCustomFieldClassCommand.php
```diff
@@ -0,0 +1,181 @@
...
+                $get->addComment('');
+            }
+            $get->addComment('@return ' . $dataType);
+
+            // If the property is a typed array, PHP will only let us use
+            // array in the return typehint.
+            $substr = substr($dataType, -2);
+            switch ($substr) {
+                case '[]':
+                    $get->setReturnType('array');
```
juampynr, line 151: This is so cool. I did not know you could create classes this way.

command/csv/field-object.csv
deviantintegral: This is a CSV copied out from mpx's documentation. See `bin/console help mpx:create-data-service` for details - it's how I bootstraped the Field class in this PR.

command/src/CreateCustomFieldClassCommand.php
deviantintegral: Indeed. Fixed!

PR #120, opened 2018-05-15
Denormalize custom fields

### LINKED ISSUE #92: Support custom fields on mpx objects

deviantintegral:
This PR resolves #92 .

The goal is straightforward; custom fields should be discoverable to developers in IDEs and treated as much as like regular fields as possible. However, there's a few complexities:

- Each object can have one or more namespaces of fields attached to it, and those namespaces may share property names. We can't just flatten them directly into an object like Media.
- Namespace prefixes may change between requests, if a specific namespace prefix has not been set with the custom fields configuration. There's no guarantee that a prefix of `pl1` is consistent between requests.
- The format mpx returns custom fields in does not match what the Symfony serializer expects. If namespace prefixes were consistent, we could implement a type extractor for them as root-level properties, but since they aren't we need to store them in a subobject. Luckily, the CJsonEncoder is the perfect spot to alter this incoming data.

I have this working with production data with custom fields; ping me if you want to see a demo of that.

TODOs:

- [x] If custom fields are set in mpx, but no class is available to contain those fields, we currently fail with an undefined index. I think we need to add a default "UnimplementedCustomFields" class to handle these cases.
- [x] Update the README with docs on how to implement custom fields.
- [x] Update the CHANGELOG, especially noting we made `deserialize` protected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 - Service URLs are now cached when loading objects and listening for
-  notifications. #144
+  notifications. #114
+- Support for loading custom fields, including denormalizing into custom
+  classes not shipped with this library. See the README for detailed
+  documentation. #120
 
 ### Changed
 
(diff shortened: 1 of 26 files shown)


juampynr:
Looks good. Merging.

PR #122, opened 2018-05-17
Discovery interface for Custom Fields

deviantintegral:
This extracts an interface so we can implement a custom discovery in the Drupal module.

### DIFF
--- a/src/DataService/CustomFieldDiscoveryInterface.php
+++ b/src/DataService/CustomFieldDiscoveryInterface.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace Lullabot\Mpx\DataService;
+
+/**
+ * Interface for classes implementing discovery of Custom Field plugins.
+ *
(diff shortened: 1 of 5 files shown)


deviantintegral:
Looks like CircleCI is down: https://status.circleci.com/incidents/q5hn2z6yfh21

src/DataService/CustomFieldDiscoveryInterface.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Lullabot\Mpx\DataService;
+
+/**
+ * Interface for classes implementing discovery of Custom Field plugins.
+ *
+ * Different frameworks and content management systems have various methods of
+ * handling discovery of annotated classes. For example, a CMS like Drupal
+ * doesn't strictly follow PSR-4, and only should expose classes for modules
```
juampynr, line 10: :+1:

PR #124, opened 2018-05-22
Make description optional for server errors

deviantintegral:
The mpx documentation is wrong and says description is included in all errors. This PR makes it required but only for client errors.

Also, we add validation that the username has a leading directory component.


juampynr:
@deviantintegral code_fixer fails:

image

I don't get what's wrong there. Can you clarify?

juampynr:
The rest of the pull request looks fine to me.

deviantintegral:
@juampynr it looks like `php-cs-fixer` had a ton of bug fixes released in the second-to-latest release. Both the yoda_style_fixer and the import fixer have bugs closed against them: https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases/tag/v2.11.2

The nightly build on master now fails too. I've pushed a commit to this branch fixing things.

PR #125, opened 2018-05-23
Add a method for setting token lifetimes

### LINKED ISSUE #23: Set a time limit for tokens

deviantintegral:
We need this API to resolve https://github.com/Lullabot/media_mpx/issues/23

### DIFF
--- a/tests/src/Unit/AuthenticatedClientTest.php
+++ b/tests/src/Unit/AuthenticatedClientTest.php
@@ -192,7 +192,7 @@ class AuthenticatedClientTest extends TestCase
     }
 
     /**
-     * @covers ::getConfig()
+     * @covers ::getConfig
      */
     public function testGetConfig()
     {
@@ -214,8 +214,8 @@ class AuthenticatedClientTest extends TestCase
     /**
      * Test setting a duration for renewed tokens.
      *
-     * @covers ::setTokenDuration()
-     * @covers ::mergeAuth()
+     * @covers ::setTokenDuration
+     * @covers ::mergeAuth
      */
     public function testTokenDuration()
     {
--- a/tests/src/Unit/AuthenticatedClientTest.php
+++ b/tests/src/Unit/AuthenticatedClientTest.php
@@ -13,6 +13,7 @@ use Lullabot\Mpx\Service\IdentityManagement\User;
 use Lullabot\Mpx\Service\IdentityManagement\UserSession;
 use Lullabot\Mpx\Tests\JsonResponse;
 use Lullabot\Mpx\Tests\MockClientTrait;
+use Lullabot\Mpx\Token;
 use Lullabot\Mpx\TokenCachePool;
 use PHPUnit\Framework\TestCase;
 use Psr\Http\Message\RequestInterface;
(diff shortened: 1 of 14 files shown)


tests/src/Unit/AuthenticatedClientTest.php
```diff
@@ -210,6 +211,32 @@ public function testGetConfig()
         $this->assertEquals('the-value', $authenticatedClient->getConfig('the-option'));
     }
 
+    /**
+     * Test setting a duration for renewed tokens.
+     *
+     * @covers ::setTokenDuration()
```
juanolalla: Is () valid syntax in @covers? I've got errors from using that

tests/src/Unit/AuthenticatedClientTest.php
```diff
@@ -210,6 +211,32 @@ public function testGetConfig()
...
+     * @covers ::mergeAuth()
```
juanolalla: I don't see how this is covering ::mergeAuth

tests/src/Unit/AuthenticatedClientTest.php
```diff
@@ -210,6 +211,32 @@ public function testGetConfig()
...
+        $userSession = $this->getMockBuilder(UserSession::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $authenticatedClient = new AuthenticatedClient($client, $userSession);
+        $duration = rand(1, 3600);
+        $userSession->expects($this->once())->method('acquireToken')
+            ->with($duration, false)
+            ->willReturn(new Token('mpx/USER-ID', 'abcdef', $duration));
+        $authenticatedClient->setTokenDuration($duration);
+        $authenticatedClient->request('GET', 'https://www.example.com/');
```
juanolalla, line 237: I don't see how this test works, no assertions?

juanolalla (commented):
I've left some comments/questions

tests/src/Unit/AuthenticatedClientTest.php
deviantintegral: Interesting, what errors? This is what phpstorm is autocompleting for me, and I haven't seen any errors come up.

tests/src/Unit/AuthenticatedClientTest.php
deviantintegral: Line 233 has the assertion through the `expects()` call. If the method is called a different number of times, or with different parameters, an assertion will fail. You can try this locally by reverting the change in the `acquireToken` method call.

tests/src/Unit/AuthenticatedClientTest.php
deviantintegral: As mergeAuth is private, it's covered through the public method call on `request`. If you generate code coverage for this test you should see mergeAuth() being called.

tests/src/Unit/AuthenticatedClientTest.php
juanolalla: covers-invalid-syntax

tests/src/Unit/AuthenticatedClientTest.php
juanolalla: :+1:

tests/src/Unit/AuthenticatedClientTest.php
juanolalla: :+1:

tests/src/Unit/AuthenticatedClientTest.php
deviantintegral: This is interesting. PHPStorm adds the `()` on autocomplete, but I can't find any documentation as to why in YouTrack. The failure you're seeing is added at `\Drupal\Tests\Listeners\DrupalStandardsListenerTrait::checkValidCoversForTest` by Drupal core, which is why there's no error in this non-Drupal library. [Remove () from a bunch of @covers definitions in PHPUnit](https://www.drupal.org/project/drupal/issues/2328919) is the closest issue I can find, which boils down to "lets follow the docs".

I filed [a bug in PHPStorm](https://youtrack.jetbrains.com/issue/WI-42254) for this.

juanolalla: Nice research, and great job overall!

PR #126, opened 2018-05-25
Add a method / interface for fetching the raw JSON response

deviantintegral:
In adding HTTP-response level caching to the Drupal module, I realized two things:

1. There was no easy way to get the response body of an object for caching.
2. We needed some way to support taking a list of results, but then caching them as if each individual item had been requested individually.

This PR enables both, by including the JSON response on each object.

Currently, I'm writing response caching in the Drupal module. However, once I'm happy with it and it hasn't changed much, I'd like to pull it into the library. That way, these JSON methods won't be needed - but will be available so downstream authors aren't ever required to do double-http requests or serialization to get the JSON representation.

### DIFF
--- a/src/DataService/DataObjectFactory.php
+++ b/src/DataService/DataObjectFactory.php
@@ -105,7 +105,7 @@ class DataObjectFactory
      * @param string $data  The JSON string to deserialize.
      * @param string $class The full class name to create.
      *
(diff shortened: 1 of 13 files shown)


juampynr:
Looks good. Tests pass. Good coverage. Merging!

PR #127, opened 2018-05-31
Optional byFields and methods for sorts and ranges

### LINKED ISSUE #78: Make $fields optional in DataObjectFactory::select()
If a developer doesn't want to apply any filters, sorts, or ranges, they should be able to execute a select() without having to add a dummy ByFields object.

deviantintegral:
Resolves #78 .


juanolalla (approved):
Good work :+1:

PR #129, opened 2018-06-06
Support Q queries

### LINKED ISSUE #128: Support Q queries
When selecting mpx objects, we should support at least basic "Q query" syntax to filter out results.

https://docs.theplatform.com/help/wsf-q-query-syntax-reference

deviantintegral:
Resolves #128.

Shockingly, it only took around 20 minutes to do the split and refactor of the ByFields class into ObjectListQuery. ❤️ test coverage and PHPStorm!

TODOs:

- [x] There are unimplemented tests which will cause the base build to fail.
- [x] Update the README and the ReadmeTest showing how to use Q queries.
- [x] Update the changelog as this _really_ breaks API compatibility!


vordude:
This looks great from what I can see. I'll wait for @juampynr 's opinion. but see nothing wrong here.

deviantintegral:
Thanks @vordude ! @juampynr feel free to leave comments for a followup.

juampynr:
Post-mergem review: looks great!

PR #131, opened 2018-06-15
Support Link and Image types

### LINKED ISSUE #130: Support Link objects
I misread the mpx documentation, and links aren't just raw URIs, but contain 'href', 'title', and so on. If custom fields define links then deserializing will fail.

https://docs.theplatform.com/help/wsf-index-of-supported-data-types#tp-toc11

deviantintegral:
This PR supports Link and Image data types, which are used in custom fields (but not in core mpx fields AFAICT). This also fixes some small bugs in the code generator for custom field classes.

https://github.com/nette/php-generator/pull/30 is the upstream issue for double-spaces between methods.

Resolves #130 .


juampynr:
Looks good. Merging.

PR #132, opened 2018-06-18
Set default values in data objects

### LINKED ISSUE #95: Set default values in data objects
Many mpx fields are optional values, such as the `programId` on `Media`. If `getProgramId()` is called, and that value hasn't been set, PHP throws a `TypeError` as it's trying to return `null`.

One option would be to [require PHP 7.1 and use a nullable return type](https://blog.shameerc.com/2016/11/php-71-8-new-features-you-need-to-know). However, that forces callers to check for `null` on all returns which is fragile, especially for object returns where you might forget and then do a method call on null.

My next idea was to return "empty" objects still satisfying the return type, such as an empty array, string, or a `Uri` object with no parameters (which results in an empty string). However, we have `\DateTime` returns, and some of those are optional. There's no good way to return an "empty" `\DateTime`.

I think this is a good place for us to use [PHP 7 TypeErrors](http://php.net/manual/en/language.errors.php7.php). It would save us from having to write `throw new PropertyNotSetException()...` in every single `get` method. However, we should double check that `getErrorCode()` lets us distinguish null errors from other types (such as returning an integer instead of a string, which would be a code bug to fix).

deviantintegral:
This resolves #95.

* All scalar types are now nullable both in their set methods and their returns. This handles both optional fields and partially loaded objects.
* New "Concrete Date" and "Null Date" classes, along with a pair of interfaces, let us handle "empty" dates.
* If you try to get a null URI object, we return a `new Uri()` which formats to an empty string.
* If you try to get a null date, you get a new Null Date object that formats to an empty string.
* It turns out all mpx objects like Releases, Chapters, and so on are always arrays. In those cases we return the default empty array.

TODOs:

- [x] Update the CHANGELOG.
- [x] Remove the expected PHP 7.0 Circle build result.


juampynr:
I love the Null pattern. I guess that this is in progress given the above CI results?

deviantintegral:
Oh yeah. Totally in progress :D

deviantintegral:
This is ready for review!

command/src/ClassGenerator.php
```diff
@@ -0,0 +1,104 @@
...
+        'String (64, ASCII)' => 'string',
+        'String (64, Unicode)' => 'string',
+        'String(64,Unicode)[]' => 'string[]',
+        'String' => 'string',
+        'TransferInfo' => '\\'.TransferInfo::class,
+        'URI' => '\\'.UriInterface::class,
+    ];
+
+    /**
+     * @param $dataType
```
juampynr: Can you add method descriptions in this class?

src/DataService/Player/Player.php
```diff
@@ -615,149 +701,173 @@ public function getAdminTags(): array
...
      */
-    public function getAlwaysShowOverlay(): bool
+    public function getAlwaysShowOverlay(): ?bool
     {
         return $this->alwaysShowOverlay;
     }
 
     /**
-     * Set Indicates whether the player should always display the play overlay.
+     * Set indicates whether the player should always display the play overlay.
```
juampynr, line 866: This and the above ones sound weird (Set indicates). I guess that this was autogenerated, right? If so, I am fine with them.

src/DataService/Player/Player.php
```diff
@@ -787,97 +897,101 @@ public function setAspectRatioHeight($aspectRatioHeight)
...
 
     /**
-     * Set URL for a custom background image.
+     * Set uRL for a custom background image.
      *
-     * @param string
+     * @param string $backgroundImageUrl
      */
-    public function setBackgroundImageUrl($backgroundImageUrl)
+    public function setBackgroundImageUrl(?string $backgroundImageUrl)
```
juampynr, line 970: Whoa! I did not know about nullable typehinting in parameters :+1:

src/DataService/Player/Player.php
```diff
@@ -887,103 +1001,103 @@ public function setColorSchemeId($colorSchemeId)
...
      */
-    public function getControlLayoutXml(): string
+    public function getControlLayoutXml(): ?string
     {
         return $this->controlLayoutXml;
     }
 
     /**
-     * Set XML layout content for the player control rack.
+     * Set xML layout content for the player control rack.
```
juampynr: heh, maybe it's better to put all in lowercase

src/DataService/Player/Player.php
```diff
@@ -1053,31 +1167,31 @@ public function getCustomJavaScriptUrls(): array
     }
 
     /**
-     * Set URLs to custom JavaScript content for the player.
+     * Set uRLs to custom JavaScript content for the player.
```
juampynr: same here

juampynr:
Just a few remarks on docblocks. Feel free to merge it.

src/DataService/Player/Player.php
deviantintegral: Yeah, this is straight from the mpx docs. It's poor reading there too, but I think it's OK to match so it's clear where it came from.

src/DataService/Player/Player.php
deviantintegral: So much for my `lcfirst` call. I'll update these and save any fixes in the class generator for the next time I need to regenerate.

PR #133, opened 2018-06-20
Rename id methods to mpxId

### LINKED ISSUE #104: Rename getId / setId to getMpxId
Other libraries are very likely to have id methods on their own objects. Since the whole point of the interface is to enable interop, we should rename the methods to be more specific to limit interface conflicts.

deviantintegral:
Resolves #104.


tests/src/Functional/ReadmeTest.php
```diff
@@ -53,7 +53,7 @@ public function testExample()
             ->setStartIndex(1)
             ->setEndIndex(1);
         $results = $mediaFactory->select($filter, $account);
-        foreach ($results as $media) {
+        foreach ($results as $index => $media) {
```
juanolalla: Why do you add $index if you don't use it within the foreach?

juanolalla (approved):
Nice work!

tests/src/Functional/ReadmeTest.php
deviantintegral: I think that was leftover from me debugging https://github.com/Lullabot/mpx-php/commit/a83d72e38d824166f6fdcfa32fefb90431495165#diff-e48dbb04bcf9bceb299a0fb63947dd48

I'll remove it!

PR #134, opened 2018-06-22
Fix mapping custom field types to arrays

deviantintegral:
This was causing the generated get / set methods to be scalars like `string` instead of `array`.


juampynr:
Looks good. Merged!

PR #135, opened 2018-07-03
Always load custom fields in responses

deviantintegral:
This comes up if a Media object not only doesn't have a custom field value, but it doesn't have any other fields set in the whole custom field namespace. That means that no custom fields object is even created for that namespace, which leads to the null method calls in code working with custom fields.

This ensures that an empty custom field object is always created.

### DIFF
--- a/src/DataService/DataObjectFactory.php
+++ b/src/DataService/DataObjectFactory.php
@@ -128,6 +128,28 @@ class DataObjectFactory
 
         $object = $this->getObjectSerializer($cached)->deserialize($data, $class, 'json');
 
+        // Adds any missing custom field classes. A given object may not
+        // contain any set fields in a given custom field namespace. In that
+        // case, the serializer will never create a field class. This is fine
+        // until calling code wants to check to see if a given field exists.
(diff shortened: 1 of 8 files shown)


src/DataService/DataObjectFactory.php
```diff
@@ -128,6 +128,28 @@ protected function deserialize($data, string $class)
...
+        // case, the serializer will never create a field class. This is fine
+        // until calling code wants to check to see if a given field exists.
+        // By ensuring we add in any missing custom field definitions, we
+        // ensure a valid object is returned instead of null.
+        if ($object instanceof ObjectInterface) {
+            $customFields = $object->getCustomFields();
+            $remaining = array_diff_key($this->dataService->getCustomFields(), $customFields);
+
+            /**
+             * @var string
```
juampynr: Should this be `@var string $namespace` ?

deviantintegral: Yes, it should!

deviantintegral: Hm. php-cs-fixer is removing that. I'll try a different syntax.

PR #140, opened 2018-07-19
Feed url support

### LINKED ISSUE #111: Support generating media feed URLs
Depends on #110 .

Documentation is at https://docs.theplatform.com/help/feeds-requesting-media-feeds. Conceptually, this is similar to generating player URLs, so we may be able to share some code.

deviantintegral:
Depends on #139 and resolves #111 .

This PR adds support for generating feed URLs. Since ID and GUIDs are not possible to include in the same URL, I've split those into two classes so it's impossible to get them mixed up. I've also added an interface shared with Player URLs for returning a rendered URI.

I thought about decorating the classes instead with `UriInterface`, but that leads to quite a bit of boilerplate without an obvious benefit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 - The FeedConfig object is now supported. #110
+- New classes have been added to support feed URL generation. See
+  `\Lullabot\Mpx\Service\Feeds` for details. #111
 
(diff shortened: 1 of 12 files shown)


aminnelson:
Does this comment count as participation?

deviantintegral:
Interesting. I can now mention you in a comment @aminnelson , but I still can't add you to the reviewer field. https://help.github.com/articles/requesting-a-pull-request-review/ says "anyone with read access" which is clearly you 😕 

I wonder what happens if you actually review the PR.

PR #141, opened 2018-08-07
Fix for player url

deepika740:
@deviantintegral 
- The player does not currently play the expected video. The generated iframe says "release with PID was not found; ignoring /select/" for the src"
- Adding /media to the url fixes the issue. 
- Our original conversation can be found here for reference --> https://github.com/Lullabot/media_mpx/issues/57

### DIFF
--- a/src/Service/Player/Url.php
+++ b/src/Service/Player/Url.php
@@ -79,7 +79,7 @@ class Url implements ToUriInterface
      */
     public function toUri(): UriInterface
     {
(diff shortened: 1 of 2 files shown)


deviantintegral:
This looks fine. I just turned on CircleCI builds for forks, and I'm guessing this test would fail: https://github.com/Lullabot/mpx-php/blob/master/tests/src/Unit/Service/Player/UrlTest.php

Can you update that and push, so we can make sure all the jobs are passing? Thanks!

m4olivei (approved):

PR #144, opened 2018-10-19
Added a script that will run php-cs-fixer

geekygnr:
This was done just so there is a reference to it being part of the process.

This was just a small little adjustment. Mostly to make my life easier as I work on this.


deviantintegral (approved):
I don't know why I never did this before. Thanks!

PR #145, opened 2018-10-25
update object interface to include more common functions

geekygnr:
This was first motivated by https://github.com/Lullabot/media_mpx/pull/72#discussion_r226679801
I did a quick  pass through the Field object and pulled all the getters and setters that looked generic to all objects.

A couple things could still either be done here or be added as an issue for future improvement
- [ ] Clean up the comments so `@inheritdoc` is being used where needed
- [ ] Refactor some of this code, I have not confirmed it but I suspect a lot of these functions are the same and could be implemented in ObjectBase instead.
- [ ] Adjust tests to match refactoring

### DIFF
--- a/src/DataService/ObjectInterface.php
+++ b/src/DataService/ObjectInterface.php
@@ -81,20 +81,6 @@ interface ObjectInterface extends IdInterface, JsonInterface
     public function setCustomFields(array $customFields);
 
     /**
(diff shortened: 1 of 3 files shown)


deviantintegral (approved):

deviantintegral:
We can do these in followups. Thanks!

PR #147, opened 2018-10-27
created a new query option to limit what fields are returned

geekygnr:
This was done to support https://github.com/Lullabot/media_mpx/issues/24 in order to get a large list of object ids.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ composer.lock
 phpunit.xml
 vendor/
 coverage/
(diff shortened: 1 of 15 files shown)


deviantintegral (commented):
Looks great!

Two things (that I may work on so we can get this merged):

- [ ] Add a functional test proving this works. The unit test is great, but it doesn't show that mpx actually respects the fields we set.
- [ ] Add a note to the CHANGELOG.md

PR #149, opened 2018-11-01
Issue #148 Make it so that a set value of FALSE for autoPlay and/or playAll is reflected in the player URL

### LINKED ISSUE #148: Player Url class does not output autoPlay=false
Mpx players can be configured to autoPlay and playAll by default. Sometimes you want to override that behaviour on a per-player basis by setting `autoPlay=false` or `playAll=false` on the query string for the player URL. The current implementation ignores a value of FALSE.

m4olivei:
Resolves #148.

### DIFF
--- a/src/Service/Player/Url.php
+++ b/src/Service/Player/Url.php
@@ -45,11 +45,11 @@ class Url implements ToUriInterface
     private $media;
 
     /**
-     * Should autoplay be overridden?
+     * Should autoPlay be overridden?
      *
      * @var bool
      */
-    private $autoplay;
+    private $autoPlay;
 
     /**
      * Should the playAll setting be overridden?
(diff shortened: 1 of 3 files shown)


geekygnr (commented):
No major issues visually.

src/Service/Player/Url.php
```diff
@@ -109,11 +109,11 @@ public function __toString()
      *
      * @see https://docs.theplatform.com/help/player-player-autoplay
      *
-     * @param bool $autoplay True to enable autoplay, false otherwise.
+     * @param bool $autoPlay True to enable autoPlay, false otherwise.
      */
-    public function setAutoplay(bool $autoplay)
+    public function setAutoplay(bool $autoPlay)
```
geekygnr, line 114: If we are following camel case should this be setAutoPlay similar to the variable autoPlay?

geekygnr: If we are following camel case should this be setAutoPlay similar to the variable autoPlay?

m4olivei: I hesitated to do that b/c that would change the public interface and be a breaking change. Plus require changing all the callers. I wanted to keep this low risk and quick.

geekygnr: Sounds good to me

PR #153, opened 2018-12-11
Support rendering URLs by GUID

deviantintegral:
I already updated the CHANGELOG, see it for details!

### DIFF
--- a/src/Service/Player/Url.php
+++ b/src/Service/Player/Url.php
@@ -123,10 +123,18 @@ class Url implements ToUriInterface
      * @see https://docs.theplatform.com/help/player-player-autoplay
      *
      * @param bool $autoPlay True to enable autoPlay, false otherwise.
+     *
+     * @return Url
      */
-    public function setAutoplay(bool $autoPlay)
+    public function withAutoplay(bool $autoPlay): self
     {
-        $this->autoPlay = $autoPlay;
+        if ($this->autoPlay == $autoPlay) {
+            return $this;
+        }
+
+        $url = clone $this;
+        $url->autoPlay = $autoPlay;
+        return $url;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Service/Player/Url.php
```diff
@@ -123,10 +138,19 @@ public function __toString()
...
-    public function setAutoplay(bool $autoPlay)
+    public function withAutoplay(bool $autoPlay): self
     {
-        $this->autoPlay = $autoPlay;
+        if ($this->autoPlay == $autoPlay) {
+            return $this;
+        }
+
+        $url = clone $this;
+        $url->autoPlay = $autoPlay;
```
deviantintegral, line 158: Apparently you can modify private properties when an object clones itself. 🍌 !

http://php.net/manual/en/language.oop5.cloning.php#122822

m4olivei: Huh, look at that.

src/Service/Player/Url.php
```diff
@@ -65,15 +68,22 @@ class Url implements ToUriInterface
...
      * Url constructor.
      *
      * @param PublicIdentifierInterface $account The account the player is owned by.
      * @param PublicIdentifierInterface $player  The player to play $media with.
-     * @param PublicIdentifierInterface $media   The media to play.
+     * @param ObjectInterface           $media   The media to play.
      */
-    public function __construct(PublicIdentifierInterface $account, PublicIdentifierInterface $player, PublicIdentifierInterface $media)
-    {
+    public function __construct(PublicIdentifierInterface $account, PublicIdentifierInterface $player, ObjectInterface $media
```
m4olivei: Is this worth a note in the CHANGELOG? If people had been passing in `\Lullabot\Mpx\DataService\PublicIdentifierInterface` objects, that wouldn't fly anymore and would throw an error right?

m4olivei (approved):
Nice change. I wondered if we were eventually going to need to support the guid form of the URL. I like the immutability of the object, seems like a good fit for this class.

Just one question left, feel free to address and merge.

PR #154, opened 2019-01-09
Set the account context when listening for notifications

deviantintegral:
When listening for notifications, there is no way to set the account context, which means that for mpx users that have multiple accounts all objects will be imported.

I ended up doing some refactoring to help limit this type of bug in the future. Before, you could pass an account context into a variety of methods. That could make for confusing bugs, if the same objects are used to load data from different accounts. Instead, I've put the account context onto the `AuthenticatedClient` itself. If you need to query with a different account context (or none at all), you would create a different client object to work with.

As well, some mpx APIs don't support account contexts and throw an error if one is provided. Since there's no way for us to know that automatically, we leave it to the developer to create a client with a `null` account context in the constructor.

If I'm lucky, this should actually reduce some of the code complexity by removing a parameter from more methods than it is added to.

TODOs:

- [x] Update the CHANGELOG

### DIFF
--- a/src/DataService/DataObjectFactory.php
+++ b/src/DataService/DataObjectFactory.php
@@ -81,17 +81,16 @@ class DataObjectFactory
     /**
      * Load a data object from MPX, returning a promise to it.
      *
-     * @param int         $id       The numeric ID to load.
-     * @param IdInterface $account
-     * @param bool        $readonly (optional) Load from the read-only service.
-     * @param array       $options  (optional) An array of HTTP client options.
+     * @param int   $id       The numeric ID to load.
+     * @param bool  $readonly (optional) Load from the read-only service.
+     * @param array $options  (optional) An array of HTTP client options.
      *
      * @return PromiseInterface
      */
-    public function loadByNumericId(int $id, IdInterface $account = null, bool $readonly = false, array $options = [])
+    public function loadByNumericId(int $id, bool $readonly = false, array $options = [])
     {
         $annotation = $this->dataService->getAnnotation();
-        $base = $this->getBaseUri($annotation, $account, $readonly);
+        $base = $this->getBaseUri($annotation, $readonly);
 
         $uri = new Uri($base.'/'.$id);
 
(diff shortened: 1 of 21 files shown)


src/DataService/DataObjectFactory.php
```diff
@@ -196,15 +199,13 @@ function (ResponseInterface $response) {
...
-     *                                         account associated with the authenticated client.
      * @param array           $options         (optional) An array of HTTP client options.
      *
      * @return ObjectListIterator An iterator over the full result set.
      */
-    public function select(ObjectListQuery $objectListQuery = null, IdInterface $account = null, array $options = []): ObjectListIterator
+    public function select(ObjectListQuery $objectListQuery = null, array $options = []): ObjectListIterator
     {
-        return new ObjectListIterator($this->selectRequest($objectListQuery, $account));
+        return new ObjectListIterator($this->selectRequest($objectListQuery, $options));
```
m4olivei, line 208: Oh, is this a bonus fix here, where `$options` wasn't being passed before?

src/DataService/NotificationListener.php
```diff
@@ -127,33 +133,51 @@ public function sync()
...
-                new UnixMillisecondNormalizer(),
-                new UriNormalizer(),
-                new ObjectNormalizer(null, null, null, $extractor),
-                new ArrayDenormalizer(),
-            ];
-
-            $serializer = new Serializer($normalizers, $encoders);
-            // @todo Handle exception returns.
-            return $serializer->deserialize($response->getBody(), Notification::class.'[]', 'json');
+            return $this->deserializeResponse($response);
```
m4olivei, line 153: 💯

m4olivei (approved):
Looks great to me. Nice work!

src/DataService/DataObjectFactory.php
deviantintegral: Yep!

juampynr:
Looks good to me too. I would like to go through this with you some day, @deviantintegral. Maybe at the retreat?

juampynr:
Actually, I meant this plus the caching layer that you added this week.

deviantintegral:
Sure thing @juampynr , you'll know where to find me!

PR #155, opened 2019-01-13
Add a command to GET an arbitrary mpx object

deviantintegral:
I was getting annoyed with hacking functional tests to load mpx objects for debugging, so I wrote a console command to do so.

```
Usage:
  mpx:get <id>
  mpx:get http://data.media.theplatform.com/media/data/Media/12345

Arguments:
  id                    The full ID of the mpx object to load.

Options:
  -h, --help            Display this help message
  -q, --quiet           Do not output any message
  -V, --version         Display this application version
      --ansi            Force ANSI output
      --no-ansi         Disable ANSI output
  -n, --no-interaction  Do not ask any interactive question
  -v|vv|vvv, --verbose  Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug

Help:
  This command loads an mpx data service object.

  An mpx username and password is required. They will be requested interactively,
  or MPX_USERNAME, and MPX_PASSWORD environment variables will be used.

  If debug logging is enabled (-vvv) then each request and response will be
  printed to the console.
```

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -3,6 +3,7 @@
 
 use Lullabot\Mpx\Command\CreateCustomFieldClassCommand;
 use Lullabot\Mpx\Command\CreateDataServiceClassCommand;
+use Lullabot\Mpx\Command\GetCommand;
 
 require_once __DIR__ . '/../vendor/autoload.php';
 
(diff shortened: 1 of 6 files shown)


m4olivei (approved):
Static review looks good. Just need a merge conflict resolution.

PR #156, opened 2019-01-24
Add new interfaces to fix wrong typehint on Url missing getPid() method

deviantintegral:
This fixes an error in calling `getPid` when generating URLs when the constructor parameter (but not the phpdocs on the private property) didn't the method. As well, this will make it simpler for Url consumers to generate URLs without having a full mpx "media" object available, since there are only three fields to implement.


m4olivei (approved):
Static review looks good. Makes sense.

PR #158, opened 2019-05-14
Allow callers of guzzle requests to retrieve the internal list object

salvamomo:
# Description

Allows callers of guzzle requests to fetch the internal DataObject list. This is related to https://github.com/Lullabot/media_mpx/pull/122, which adds a UI form to queue items through batches, and having the total count of items at the batch processing level is rather convenient.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased] - YYYY-MM-DD
 
+### Added
+
+- `\Lullabot\Mpx\DataService\ObjectListIterator.php` now allows the caller to
+  get the total number of mpx results, via `::getTotalResults()` method.
+
 ### Changed
 
 - New interfaces have been added, which ObjectInterface now inherits from.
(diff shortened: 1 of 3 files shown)


deviantintegral (commented):
My initial thought is that it would be safer to proxy the value by adding `getTotalResults()` on `ObjectListIterator`. That way, we don't have to worry about callers modifying the data for a single page while the iterator may be in the middle of a loop. Thoughts?

salvamomo:
@deviantintegral Agreed. I had added something similar initially in that class, in fact, and removed thinking that maybe it wasn't that suitable for the Iterator class... although on second thought, better not let anyone mess up with the current data while it's being processed 👍 

You'll want to squash the commits into one, as I fooled around a bit, not knowing for sure what phpcs was complaining about.

deviantintegral (approved):

deviantintegral:
No squashing, because rebasing repositories except for security issues is 😈 .

Can you update `CHANGELOG.md` too? Then feel free to merge this.

salvamomo:
Didn't notice the changelog! I really like the keep a changelog format btw, it's what I use for my time tracking app, too \o/. Merging now.

PR #165, opened 2020-06-02
Fix test failures related to CachingPhpDocExtractor code

m4olivei:
Relevant upstream issue: https://github.com/symfony/symfony/issues/36049

Relevant upstream pull request that I'm basing the changes on: https://github.com/symfony/symfony/pull/36850/files

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,2 +1,3 @@
 exclude_patterns:
   - "src/DataService/CachingPhpDocExtractor.php"
+  - "tests/"
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,2 @@
+exclude_patterns:
+  - "src/DataService/CachingPhpDocExtractor.php"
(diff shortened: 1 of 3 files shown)


m4olivei:
Arg. PHP 7.1 gets v3.1.0 of `phpdocumentor/reflection-docblock` and PHP >= 7.2 gets v5.1.0. Thats the reason for the discrepancy. Correcting it for one breaks the other. I'm going to take a different tact, which is to just remove the problem test case, and leave the code as is.

m4olivei:
Given that the class failing codeclimate is `\Lullabot\Mpx\DataService\CachingPhpDocExtractor`, which is copy pasta from Symfony to overcome some `private` vars we needed to override, I'm not willing to address the cognitive complexity issue in it. @deviantintegral could we set the exception in Code Climate there?

See https://github.com/Lullabot/mpx-php/pull/89#discussion_r180483516 for reference.

deviantintegral:
Ugh, that version split is frustrating but understandable. I think it's OK to remove the test given it's copied from upstream. It was mostly me being overcautious.

Also fine with not refactoring the Syfmony code, we can do that with the usual phpcs ignore flags I think. If you felt like refactoring and fixing upstream (assuming it still exists), that'd be a nice bonus, but otherwise 🤷 .

m4olivei:
@deviantintegral Hum, adding the phpcs ignore flags didn't phase code climate. In the past I remember you could add exceptions in the code climate UI. I don't appear to have acees to do that for this repo :/

deviantintegral:
Of course, this is not a style check!

Looks like you can configure this in a yaml file: https://docs.codeclimate.com/docs/excluding-files-and-folders

deviantintegral (approved):

PR #167, opened 2021-02-08
Adjust tests to start from a new Url instance each time, fix double equal bug

m4olivei:
Was seeing an issue where the `autoPlay=false` query string param wouldn't show in the player URL generated. Odd b/c there was a test that seemed to show it was working fine. Turns out the test setup was showing a false positive.

### DIFF
--- a/src/Service/Player/Url.php
+++ b/src/Service/Player/Url.php
@@ -148,7 +148,7 @@ class Url implements ToUriInterface
      */
     public function withAutoplay(bool $autoPlay): self
     {
-        if ($this->autoPlay == $autoPlay) {
+        if ($this->autoPlay === $autoPlay) {
             return $this;
         }
 
@@ -167,7 +167,7 @@ class Url implements ToUriInterface
      */
     public function withPlayAll(bool $playAll): self
     {
-        if ($this->playAll == $playAll) {
+        if ($this->playAll === $playAll) {
             return $this;
         }
 
@@ -182,7 +182,7 @@ class Url implements ToUriInterface
      */
     public function withEmbed(bool $embed): self
     {
-        if ($this->embed == $embed) {
+        if ($this->embed === $embed) {
             return $this;
         }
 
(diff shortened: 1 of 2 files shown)


src/Service/Player/Url.php
```diff
@@ -148,7 +148,7 @@ public function __toString()
      */
     public function withAutoplay(bool $autoPlay): self
     {
-        if ($this->autoPlay == $autoPlay) {
+        if ($this->autoPlay === $autoPlay) {
```
deviantintegral, line 151: I don't see a test covering this specifically? I would have expected `$this->autoPlay` to somehow be set to a non-boolean.

deviantintegral (approved):
A question, but you're more in touch this this code than I am. Merge at will!

src/Service/Player/Url.php
m4olivei: The trouble was that when `$this->autoPlay === NULL`, when you try to set `$url->withAutoplay(FALSE)` this conditional with a double equal would evaluate to `TRUE` and then `$this->autoPlay` would remain as `NULL`. The tests as they were were hiding this bug b/c the tests would build on a single instance of a `Url` object, first testing `$url->withAutoplay(TRUE)`, which would indeed set the value for `$this->autoPlay`, then follow that assertion by setting `$url->withAutoplay(FALSE)`, which would indeed set the value for `$this->autoPlay` to `FALSE`, behaving as expected.

In short, this would work:

```
$url = new \Lullabot\Mpx\Service\Player\Url();
$url = $url->withAutoPlay(TRUE);
$url = $url->withAutoPlay(FALSE);
$url_string = (string) $url;
/// $url_string has &autoPlay=false
```
But this wouldn't work:
```
$url = new \Lullabot\Mpx\Service\Player\Url();
$url = $url->withAutoPlay(FALSE);
$url_string = (string) $url;
/// $url_string is missing &autoPlay=false. sad panda.
```
I changed the tests so that each assertion acts on a new instance.

PR #168, opened 2021-02-09
Try matching up the xdebug paths

m4olivei:


m4olivei:
I did see the job succeed once here. Then I tested fate and tried to run it again with SSH and it failed :cry: . Seems like CircleCI is having a bad day. I'll come back here later and kick it off again.

.circleci/config.yml
```diff
@@ -78,7 +78,7 @@ test_and_cover: &test_and_cover
       - run: composer update -n --prefer-dist
 
       - run: |
-          [ -f /usr/local/lib/php/extensions/no-debug-non-zts-20200930/xdebug.so] || pecl install xdebug
+          [ -f /usr/local/lib/php/extensions/no-debug-non-zts-20190902/xdebug.so] || pecl install xdebug
```
deviantintegral, line 81: ~~It looks like I typo'ed this in a few projects.~~ What I'd expect is that this would only be a suboptimal use of the cache (since this will always run), not a segfault.

Here's another example that did pass: https://github.com/deviantintegral/jms-serializer-uri-handler/pull/1/files

m4olivei: That makes sense. I was totally guessing with this. Bummer that it's so intermittent at the moment. Probably just need to ignore it for a bit.

m4olivei:
I forgot about this, but I think I found a reliable way to solve for it over here https://github.com/Lullabot/mpx-php/pull/172/commits/3ea8500976d9fbf49a30bc27fdedc19206f6c6a1. After some Google'ing I found https://bugs.xdebug.org/view.php?id=1729 which suggests to uninstall any xdebug module first before enabling a new version.

PR #172, opened 2021-03-24
171: Update symfony depencencies

m4olivei:
Updates Symfony dependencies to allow for ^4.4. We're pretty well compatible without any change, there was only issues with test cases and one minor method signature change.

Symfony 5.5 will be another story unfortunately. There are a couple places where we're using code deprecated in 4.x and removed in 5.x. That unfortunately means that we will won't be able to allow for all three major branches, at least that's the case with `symfony/lock` specificially.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,7 +78,7 @@ test_and_cover: &test_and_cover
       - run: composer update -n --prefer-dist
 
       - run: |
-          [ -f /usr/local/etc/php/conf.d/xdebug.ini] || rm /usr/local/etc/php/conf.d/xdebug.ini
+          rm -f /usr/local/etc/php/conf.d/xdebug.ini
 
       - run: |
           [ -f /usr/local/lib/php/extensions/no-debug-non-zts-20200930/xdebug.so] || pecl install xdebug
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,6 +78,9 @@ test_and_cover: &test_and_cover
       - run: composer update -n --prefer-dist
 
       - run: |
+          [ -f /usr/local/etc/php/conf.d/xdebug.ini] || rm /usr/local/etc/php/conf.d/xdebug.ini
+
+      - run: |
           [ -f /usr/local/lib/php/extensions/no-debug-non-zts-20200930/xdebug.so] || pecl install xdebug
           echo 'zend_extension=/usr/local/lib/php/extensions/no-debug-non-zts-20190902/xdebug.so' > /usr/local/etc/php/conf.d/xdebug.ini
           echo 'xdebug.mode="coverage"' >> /usr/local/etc/php/conf.d/xdebug.ini
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+## Changed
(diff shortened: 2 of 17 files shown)


.circleci/config.yml
```diff
@@ -77,6 +77,9 @@ test_and_cover: &test_and_cover
 
       - run: composer update -n --prefer-dist
 
+      - run: |
+          [ -f /usr/local/etc/php/conf.d/xdebug.ini] || rm /usr/local/etc/php/conf.d/xdebug.ini
```
m4olivei: I had to do this b/c I was getting a segfault on enabling the xdebug extension below. This disables the existing xdebug extension before downloading and enabling the latest xdebug from pecl below.

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+## Changed
+
+- Updated Symfony dependencies to allow for ^4.4.
```
m4olivei, line 11: @deviantintegral I think this qualifies as a minor update?

.circleci/config.yml
deviantintegral: Makes sense!

You can simplify this with `rm -f`:

```
❯ rm /tmp/asdf; echo $?
rm: /tmp/asdf: No such file or directory
1
❯ rm -f /tmp/asdf; echo $?
0
```

CHANGELOG.md
deviantintegral: Yep, as long as the old version still works!

composer.json
```diff
@@ -29,7 +29,11 @@
     "symfony/console": "^3.4",
     "namshi/cuzzle": "^2.0",
     "rtheunissen/guzzle-log-middleware": "^0.4.1",
-    "mockery/mockery": "^1.0"
+    "mockery/mockery": "^1.0",
+    "symfony/phpunit-bridge": "5.2.x-dev#f2f94fd78379cdcdef09dd5025af791301913968",
```
deviantintegral, line 33: Curious why we're pinning this?

phpunit.xml.dist
```diff
@@ -20,6 +20,9 @@
              Note this will EXPOSE PASSWORDS AND TOKENS in logs, so
              this should only be enabled on local environments. -->
         <env name="MPX_LOG_CURL" value="false" />
+        <!-- Watch for deprecations, but only fail on direct deprecations,
+             not indirect ones. -->
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[direct]=0" />
```
deviantintegral, line 25: Nice find!

src/DataService/NotificationTypeExtractor.php
```diff
@@ -30,7 +30,7 @@ public function setClass(string $class)
     /**
      * {@inheritdoc}
      */
-    public function getTypes($class, $property, array $context = [])
+    public function getTypes($class, $property, array $context = []): ?array
```
deviantintegral, line 33: This technically represents an API break. In retrospect, I think we should have marked some classes like this `@internal`.

Given we don't know of a single external use of this library, and that I'm guessing that returning a non-array would break things anyways, I think we can document this in the changelog instead of forcing a whole new major version.

deviantintegral (approved):
Nice! 👏

composer.json
m4olivei: Honestly I don't remember. I feel like I would have either a) had a good reason, or b) read it somewhere as recommended for this package. I'm going to go ahead and trust my former self on this one and leave it be :smile:

PR #174, opened 2021-07-10
Issue #173 Tweak annotations to avoid PHP notice in PHP 7.4

### LINKED ISSUE #173: Add workaround to allow tests to pass on PHP 7.4 with doctrine/annotations 1.4.0
While working on Drupal 9 compatibility for `drupal/media_mpx`, I found test failures stemming from an error that only pops up when running PHP 7.4 with `doctrine/annotations:1.4.0`. There is an array notice that breaks tests. `drupal/core-recommended` pins to `doctrin/annotations:1.4.0`, so we can't update that to a more recent version that has a fix (see https://github.com/doctrine/annotations/issues/273). 

Let's workaround it by tweaking the annotations from this package that it's choking on.

m4olivei:
Resolves #173

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
-- Changed the annotations that ended in @endcode to have some trailing text to
+- Changed the annotations that ended in @endcode to have some trailing docs to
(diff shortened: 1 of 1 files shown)


m4olivei:
Blarhg, I branched wrong and ended up merging the commits intended here into master over on https://github.com/Lullabot/mpx-php/pull/175. Oh well, alls well that ends well.

PR #179, opened 2022-03-09
Issue #178 ssl substitution

### LINKED ISSUE #178: Support ssl substitution in defaultThumbnailUrl property of Media object
The `defaultThumbnailUrl` property of `Media` objects can include dynamic elements known as substitutions per the [MPX documentation](https://docs.theplatform.com/help/media-delivery-urls).

I've just run into the `ssl` substitution, where the `defaultThumbnailUrl` value looks like this:

`{ssl:https:http}://tve-static-nbc.nbcuni.com/prod/image/338/734/VCE_S21_ArianaAdvice_Thumbs_800x450_1953495619509.jpg`

The `ssl` substitution is [described here](https://docs.theplatform.com/help/media-delivery-urls#tp-toc38).

> Specifies a URL segment to use for http and https URLs, when they differ. For SSL, the first string is substituted and for non-SSL, the second string is used.

Lets account for the `ssl` substitution by replacing it per the above description.

m4olivei:
Resolves #178

### DIFF
--- a/src/DataService/Media/Media.php
+++ b/src/DataService/Media/Media.php
@@ -173,7 +173,7 @@ class Media extends ObjectBase implements PublicIdWithGuidInterface
     /**
      * The streamingUrl of the default thumbnail for this Media.
      *
-     * @var Uri
+     * @var string
      */
     protected $defaultThumbnailUrl;
 
@@ -811,8 +811,14 @@ class Media extends ObjectBase implements PublicIdWithGuidInterface
 
     /**
      * Returns the streamingUrl of the default thumbnail for this Media.
+     *
+     * Note that the defaultThumbnailUrl may include dynamic URL substitution
+     * parameters. Use getDefaultThumbnailUri() to get a Uri object with
+     * dynamic URL substitution parameters resolved.
+     *
+     * @see https://docs.theplatform.com/help/media-delivery-urls
      */
-    public function getDefaultThumbnailUrl(): Uri
+    public function getDefaultThumbnailUrl(): string
     {
         return $this->defaultThumbnailUrl;
     }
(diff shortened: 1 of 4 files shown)


src/DataService/Media/Media.php
```diff
@@ -811,6 +811,12 @@ public function setCredits(array $credits)
 
     /**
      * Returns the streamingUrl of the default thumbnail for this Media.
+     *
+     * Note that the defaultThumbnailUrl may include dynamic URL substitution
```
deviantintegral, line 815: That's amazing, because I'm pretty sure that breaks the RFC for what a valid URL protocol is like! It's too bad they didn't go with protocoless URLs.

tests/src/Unit/DataService/Media/MediaTest.php
```diff
@@ -63,6 +64,27 @@ public function testInstanceOf($field, $class)
...
+     * Test the getNormalizedDefaultThumbnailUrl method.
+     *
+     * The getNormalizedDefaultThumbnailUrl method is unlike the others in that
+     * it does not have a direct property mapping in the MPX data.
+     */
+    public function testGetNormalizedDefaultThumbnailUrl()
+    {
+        // Test against the default media fixture loaded in the setup method.
+        $object = $this->deserialize($this->class, 'defaultThumbnailUrl');
+        $this->assertEquals(new Uri('ftp://winnas01.theplatform.com/Content/Video/thePlatform_Documentation_-_SA/3/494/ian_1_flexible_320x240.jpg'), $object->getNormalizedDefaultThumbnailUrl());
```
deviantintegral, line 77: 😱 ftp!

deviantintegral (commented):
Nice work! 🙌

tests/src/Unit/DataService/Media/MediaTest.php
m4olivei: I know right?! I wondered where you got that test object from when I saw this :laughing:

src/DataService/Media/Media.php
m4olivei: Yeah :shrug:

PR #185, opened 2022-10-12
Add support for PHP 8.0 and 8.1

chrissnyder2337:
- Updates Phpunit to versions that support PHP8
- Adds a dependency on the PHP cache adapter directly and removes the dependency on `cache/cache`. `cache/cache` adds all adapters, which may not be needed.
    - `cache/cache` currently does not have a stable release for PHP >= 8.0.
    - `cache/array-adapter` (the only adapter I could find in use in the library) does support the latest versions of PHP.
    
    
Motivation: PHP 7.4 will become EOL and stop receiving security support on Nov 28, 2022.

Note: There may be additional modifications needed to support PHP >= 8.0 . These changes mainly focused on modifications needed to install the package when using PHP 8.1.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,14 +13,15 @@
     "guzzlehttp/guzzle": "^6.3",
     "guzzlehttp/psr7": "^1.0",
     "psr/log": "^1.0",
-    "cache/cache": "2.0.x-dev#83fabc5bad8b90f7981c97898345df8935b0bf55 as 1.0",
     "symfony/serializer": "^3.4 || ^4.4",
     "symfony/property-access": "^3.4 || ^4.4",
     "doctrine/annotations": "^1.2",
     "symfony/property-info": "^3.4 || ^4.4",
     "symfony/lock": "^3.4 || ^4.4",
     "symfony/finder": "^3.0 || ^4.0 || ^5.0",
-    "phpdocumentor/reflection-docblock": "^5.2"
+    "phpdocumentor/reflection-docblock": "^5.2",
+    "cache/array-adapter": "^1.0",
+    "psr/http-message": "^1.0"
   },
   "require-dev": {
     "phpunit/phpunit": "^6.5||^7.0",
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     "guzzlehttp/psr7": "^1.0",
     "psr/log": "^1.0",
     "symfony/serializer": "^3.4 || ^4.4",
-    "symfony/property-access": "^3.4 || ^4.4",
+    "symfony/property-access": "^4.4",
     "doctrine/annotations": "^1.2",
     "symfony/property-info": "^3.4 || ^4.4",
     "symfony/lock": "^3.4 || ^4.4",
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     "symfony/serializer": "^3.4 || ^4.4",
     "symfony/property-access": "^3.4 || ^4.4",
     "doctrine/annotations": "^1.2",
-    "symfony/property-info": "^4.4",
+    "symfony/property-info": "^3.4 || ^4.4",
     "symfony/lock": "^3.4 || ^4.4",
     "symfony/finder": "^3.0 || ^4.0 || ^5.0",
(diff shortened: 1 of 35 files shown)


composer.json
```diff
@@ -13,17 +13,18 @@
...
-    "cache/cache": "2.0.x-dev#83fabc5bad8b90f7981c97898345df8935b0bf55 as 1.0",
     "symfony/serializer": "^3.4 || ^4.4",
     "symfony/property-access": "^3.4 || ^4.4",
     "doctrine/annotations": "^1.2",
     "symfony/property-info": "^3.4 || ^4.4",
     "symfony/lock": "^3.4 || ^4.4",
     "symfony/finder": "^3.0 || ^4.0 || ^5.0",
-    "phpdocumentor/reflection-docblock": "^5.2"
+    "phpdocumentor/reflection-docblock": "^5.2",
+    "cache/array-adapter": "^1.0",
```
m4olivei, line 23: Oh, this is great, I like that a lot. I wonder why we didn't do this originally. I too can only find uses of `Cache\Adapter\PHPArray\ArrayCachePool`. Not seeing any other adapters in use in the package. Any recollections @deviantintegral ?

tests/src/Unit/DataService/FieldTest.php
```diff
@@ -47,7 +47,7 @@ public function testGetSet(string $field, $expected = null)
     public function getSetMethods()
     {
         $tests = parent::getSetMethods();
-        $tests['added'] = ['added', new \Lullabot\Mpx\DataService\DateTime\ConcreteDateTime(\DateTime::createFromFormat('U.u', '1236030615.000'))];
+        $tests['added'] = ['added', new ConcreteDateTime(\DateTime::createFromFormat('U.u', '1236030615.000'))];
```
m4olivei, line 50: :nail_care:

m4olivei (changes requested):
Thanks for the contribution! Conceptually all these changes make good sense to me. I'd like us to get to a clean test run before we merge this, but static review all looks nice.

I'm going to spend a little bit of time this afternoon on this to see if I can push this forward some if any.

composer.json
deviantintegral: I think I just wasn't aware that split packages were provided. TIL'ed!

m4olivei (approved):
I'm happy with this now. The code climate job is being unnecessarily pedantic IMO. I adjusted settings so it'll no longer fail the job.

@chrissnyder2337 Does this work for you?

chrissnyder2337:
This is great! With this change, I was able to install the media_mpx Drupal module on a Drupal 9.4 site running PHP 8.1.

I used this fork by adding the following to my project's composer repositories config and then running `composer require drupal/media_mpx:^2.0 lullabot/mpx-php:"dev-php-8.1 as 1.1.0"`

```
"mpx-php": {
    "type": "vcs",
    "url": "https://github.com/chrissnyder2337/mpx-php.git"
},
 ```

 
With this change I was able to:

- Set up the MPX account 
- Add the MPX users
- Set up the Media Type in Drupal.

However, I could not import an MPX media item into Drupal. However, this issue may not be related to the changes here but maybe another issue or an issue with the Drupal module's implementation. I haven't yet traced down the root cause of the error I'm seeing. I have not used this library/module before so not sure if it is a regression. The error I am seeing when trying to import an MPX media item: 
> Uncaught PHP Exception Symfony\Component\Serializer\Exception\NotNormalizableValueException: "The type of the "defaultThumbnailUrl" attribute for class "Lullabot\Mpx\DataService\Media\Media" must be one of "Lullabot\Mpx\DataService\Media\Uri" ("string" given)." at /var/www/vendor/symfony/serializer/Normalizer/AbstractObjectNormalizer.php line 533`

m4olivei:
@chrissnyder2337 I believe we ran into a similar error and resolved that over here: https://github.com/Lullabot/mpx-php/pull/188

We're using media_mpx module with this library on Drupal 9.4 / PHP 7.4. We're currently getting all the things ready for PHP 8.1 so actively fixing things like that.

I'm going to merge this and tag a release. Thanks for the contribution!

PR #188, opened 2022-10-21
Update DataService Uri vars to fully qualified names

rokit88:
Some DataService var namespaces (Uri and UriInterface) aren't resolved properly (by PhpDocExtractor I believe) without fully qualified names.


m4olivei (approved):
LGTM

PR #189, opened 2023-04-20
feat: allow installing Guzzle 7

e0ipso:
In our effort to upgrade our dependencies, we would like to allow Guzzle 7.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
   "license": "MIT",
   "require": {
     "php": ">=7.4",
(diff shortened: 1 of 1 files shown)


marcoscano (approved):
LGTM!

deviantintegral:
If you plan on continuing to support Guzzle 6, it may be worth adding an additional test job that runs composer with `--prefer-lowest`. On other projects, I've done this in whatever job tests the oldest supported PHP version.

https://github.com/deviantintegral/har/blob/main/.circleci/config.yml#L205-L214

PR #190, opened 2023-04-20
fix: cuzzle is a soft & optional dependency

e0ipso:
We should be careful using the classes defined in `namshi/cuzzle`. I am not familiar with the whole code base. It is possible that `MpxCommandBase` is only used in a `dev` context.

### DIFF
--- a/command/src/MpxCommandBase.php
+++ b/command/src/MpxCommandBase.php
@@ -65,18 +65,10 @@ abstract class MpxCommandBase extends Command
 
         $config = Client::getDefaultConfiguration();
 
-        $cl = new ConsoleLogger($output);
-
-        // Cuzzle is a soft & optional dependency. Its classes may not exist.
-        if (class_exists('Namshi\Cuzzle\Middleware\CurlFormatterMiddleware')) { // phpcs:ignore
-            /** @var $handler \GuzzleHttp\HandlerStack */
-            $handler = $config['handler'];
-            $handler->after('cookies', new Namshi\Cuzzle\Middleware\CurlFormatterMiddleware($cl)); // phpcs:ignore
-        }
-
-        $responseLogger = new Logger($cl);
+        $responseLogger = new Logger(new ConsoleLogger($output));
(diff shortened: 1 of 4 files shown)


e0ipso:
@marcoscano as you can see in https://github.com/Lullabot/mpx-php/blob/main/composer.json#L31 `namshi/cuzzle` is a require-dev dependency.

m4olivei:
@e0ipso this seems like a safe change to me. 

I don't believe these commands are used in any sort of day-to-day function. I've honestly never used them. They look like they are used for debugging. In any case, as all cuzzle is doing is logging out a curl request, it's strictly debugging and seems to me to be safe to remove. I like the soft dependency approach, but if you felt like going further, and taking it out altogether, that would be fine as well. Up to you.

deviantintegral:
> It is possible that MpxCommandBase is only used in a dev context.

This is correct - just tests and the console command, which is only available if this is the root project.

cuzzle is a library which honestly is pretty useful in the debugging cycle. You can replay requests when they break just by copying and pasting the curl command.

There's two places it's used:

https://github.com/Lullabot/mpx-php/blob/d2055dc7cd058f5321c5a1343447a78f1f85fbee/tests/src/Functional/FunctionalTestBase.php#L65-L77

which is configured in https://github.com/Lullabot/mpx-php/blob/5d2437f1aeca36fe507f36e07f683bfe6edae0ae/phpunit.xml.dist#L19-L22

And then the command class you already found in this PR. I think that's only used in the get command (again, used for debugging), and only if you pass `-vvv`:

https://github.com/Lullabot/mpx-php/blob/d3748bdf54141dce37abd17e42d678ccac76cbd6/command/src/GetCommand.php#L26-L27

I'm OK with removing the dependency if the upstream library has been abandoned. I had also created a Drupal module for it at https://drupal.org/project/cuzzle but neither seemed to get much use. I think I probably overestimated how useful this type of debugging output would be.

m4olivei (approved):
LGTM!

PR #191, opened 2023-04-21
fix: make return types consistent with parent

e0ipso:


e0ipso:
Self merging to unblock progress. This PR should be pretty simple.

PR #192, opened 2023-04-21
feat: upgrade to PHP 8.1 using rector

e0ipso:
BREAKING CHANGE: This version is no longer compatible with PHP7.4. We'll be tagging 2.0.0 with this commit.

This upgrade took an unreasonable amount of effort. Some of the highlights:

- Updated the CircleCI base image to use PHP8.1. This image didn't have some packages that were available in the previous image that we needed: imagik. This package is no longer accessible via docker-pecl-extension, so we had to do it manually.
- Updated the codebase using Rector.
- Updated the Symfony version, and find all the breaking changes. Thankfully there is an amazing test coverage that sprung all the BC break booby traps.
- Updated all the tests to make them compatible with all the BC breaks of PHPUnit versions.
- CS-Fixer also had breaking changes.
- Several dependencies had inter-locking combinations due to upgrade to PHP8.1.
- Access to Code Climate was impossible to resolve, so I decided to stop updating coverage info there, and upload the HTML report to CicleCI artifacts instead.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -143,7 +143,7 @@ code_fixer: &code_fixer
           key: v2-test-dependencies-{{ .Environment.CIRCLE_JOB }}-{{ checksum "composer.json" }}-{{ checksum ".circleci/config.yml" }}
 
       # Run tests!
-      - run: PHP_CS_FIXER_IGNORE_ENV=1 vendor/bin/php-cs-fixer fix --config=.php-cs-fixer.dist.php -v --dry-run --using-cache=no --path-mode=intersection -- src tests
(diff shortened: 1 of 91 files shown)


e0ipso:
@marcoscano for some reasons GitHub is not getting the correct integration with CircleCI in the PR. However, tests are working if you log in to CircleCI.

marcoscano (approved):
@e0ipso are there breaking changes in the API, or is it just incompatible dependencies?
If we have changes in the API, it would be great to have them documented in the Readme (or somewhere else), on top of tagging a new major with this.
Apart from that, 👍 from me!

e0ipso:
@marcoscano the library does not have API changes, only infrastructure and dependency changes.

PR #197, opened 2023-10-04
feat: D10 compatibility

marcoscano:
Let's create a new major bumping all dependencies to be D10-only.


src/Normalizer/UnixMillisecondNormalizer.php
```diff
@@ -5,24 +5,69 @@
...
 /**
  * Denormalize a millisecond timestamp into a date object.
+ *
+ * Note this doesn't extend the
+ * \Symfony\Component\Serializer\Normalizer\DateTimeNormalizer normalizer
+ * directly because the ::denormalize() method in that class is typehinted to
+ * \DateTimeInterface and we want to return a ConcreteDateTime object instead.
+ * Unfortunately \DateTimeInterface can't be extended by user classes, so the
+ * option we have is to unfortunately copy here most methods from
+ * DateTimeNormalizer, and modify the ones that are particular to us here.
```
marcoscano: This is unfortunately the only way I could find to workaround the fact that we can't extend `\DateTimeInterface` but I would be happy to go into another direction if others can think of alternative approaches.

e0ipso: Would it be possible to use a [decorator pattern](https://en.wikipedia.org/wiki/Decorator_pattern) here? I think this is a perfect use case for it.

src/Normalizer/UriNormalizer.php
```diff
@@ -58,11 +49,13 @@ public function denormalize($data, $class, $format = null, array $context = [])
...
-     */
     public function supportsDenormalization($data, $type, $format = null)
     {
         return isset(self::$supportedTypes[$type]);
     }
+
+    public function getSupportedTypes(?string $format): array
+    {
+        return self::$supportedTypes;
+    }
```
e0ipso, line 60: Is this a new part of the interfaces?

tests/src/Unit/AuthenticatedClientTest.php
```diff
@@ -310,36 +310,9 @@ private function fetchTokenLogger(int $count)
...
-                                '!\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\+\d{4}!',
-                                $context['date']
-                            );
-                        } catch (ExpectationFailedException) {
-                            return false;
-                        }
-
-                        return true;
-                    })]);
+                    ['Expiration defined for "{resource}" lock for "{ttl}" seconds.']);
```
e0ipso, line 315: As I understand it, these two log events were moved from level `info` to level `debug` in Symfony 6.

Additionally I see we are dropping the test for our own logging, which happens with `info` level. Should we keep that test?

My preferred course of action would be to move our logging to `debug` and then make a minor tweak in the test:

```diff
- $logger->expects($this->any())->method('info')
+ $logger->expects($this->any())->method('debug')
```

tests/src/Unit/Service/IdentityManagement/UserSessionTest.php
```diff
@@ -117,38 +117,14 @@ public function testAcquireReset()
...
-                        $this->assertMatchesRegularExpression('!\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\+\d{4}!', $context['date']);
-                    } catch (ExpectationFailedException) {
-                        return false;
-                    }
-
-                    return true;
-                })]
-            );
+                ['Expiration defined for "{resource}" lock for "{ttl}" seconds.']);
+        $logger->expects($this->any())->method('info')
```
e0ipso, line 125: Similar to the comment above, if we log this event with `debug` level, test changes simplify a lot.

src/Normalizer/UriNormalizer.php
marcoscano: Yes, this is now a required method of the normalizers. 👍

tests/src/Unit/AuthenticatedClientTest.php
marcoscano: Sorry git is making this look more convoluted than it is. 
I did change the two messages that are now `debug` into a new `::method()` call, but then also simplified the remaining `info` call to drop the callback, since now the strings are being compared as-is.

src/Normalizer/UnixMillisecondNormalizer.php
marcoscano: Good call! I'll give it a try. Thanks!

tests/src/Unit/AuthenticatedClientTest.php
e0ipso: > simplified the remaining info call to drop the callback, since now the strings are being compared as-is

Are we losing test coverage with this change?

marcoscano: A little bit, but I don't think it's too significant (AFAICT). The callback was asserting that those particular tokens were part of the context passed in, and that if there was a `{date}` token, it matched a specific regular expression. It seemed to me that asserting that the (tokenized) message is indeed logged when expected was good enough. I can spend more time on it trying to troubleshoot if you have strong opinions on maintaining the callback checks though. Just let me know!

src/Normalizer/UnixMillisecondNormalizer.php
```diff
@@ -5,24 +5,65 @@
...
+    }
+
+    public function setDefaultContext(array $defaultContext): void
+    {
+        $this->decorated->setDefaultContext($defaultContext);
+    }
+
+    public function getSupportedTypes(?string $format): array
+    {
+        return self::$supportedTypes;
```
e0ipso, line 50: Ugh! If you don't override here it won't work due to _late static binding_, right? I wish the decorated used `static` instead of `self`.

tests/src/Unit/AuthenticatedClientTest.php
e0ipso: got it! You are right, we shouldn't need to test upstream implementations.

e0ipso (approved):
Let's merge!

PR #198, opened 2023-10-11
feat: D10 compatibility follow-up 1

marcoscano:


marcoscano:
Nice work @e0ipso ! Thanks for picking this up and refactoring the cache adapter. I want to approve this but I apparently can't, so here's my 👍 on the PR.

PR #199, opened 2024-02-19
Call to undefined method Symfony\Component\Cache\Adapter\Psr16Adapter::set()

dzmitrykruchko-nbcuni:


marcoscano (approved):
This looks good to me. I wonder if there are other places in the library that are no longer compatible, but I think we'll end up finding out as we move forward. I am OK doing it iteratively.

There is a failure in CircleCI code inspections but I am not sure if they come from this PR, maybe we can try to fix it before merging if that's the case?

The command being run in CI is:

```
vendor/bin/php-cs-fixer fix --config=.php-cs-fixer.dist.php -v --dry-run --using-cache=no --path-mode=intersection -- src tests
```

dzmitrykruchko-nbcuni:
> This looks good to me. I wonder if there are other places in the library that are no longer compatible, but I think we'll end up finding out as we move forward. I am OK doing it iteratively.
> 
> There is a failure in CircleCI code inspections but I am not sure if they come from this PR, maybe we can try to fix it before merging if that's the case?
> 
> The command being run in CI is:
> 
> ```
> vendor/bin/php-cs-fixer fix --config=.php-cs-fixer.dist.php -v --dry-run --using-cache=no --path-mode=intersection -- src tests
> ```

The tests issue is not related to the changes in this PR as `Found 45 of 178 files that can be fixed`
We can apply CS Fixer fixes here, but i think it makes sense to do it in a separate PR.

PR #200, opened 2024-02-20
Apply CS fixes

deviantintegral:
I would have expected this to fail the PR where things were upgrade. It's surprising, I no longer see any job data for PRs like this one: https://github.com/Lullabot/mpx-php/pull/192

I'm not sure I like removing the fqcn's from docblocks on imported classes, but I don't see an easy way to disable that and if Symfony is doing it we may as well follow.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ coverage/
 /.idea
 /build
 .phpunit.result.cache
+.php-cs-fixer.cache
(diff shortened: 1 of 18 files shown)


deviantintegral:
Well this fixed those tests, but PHP builds are failing. We need the FQCN's for the serializer.

marcoscano (approved):
LGTM!

deviantintegral:
@marcoscano any idea what's up with the missing expected CI reports?

e0ipso:
Closed in favor of https://github.com/Lullabot/mpx-php/pull/202

PR #201, opened 2025-09-29
upgrade psr/http-message

dzmitrykruchko-nbcuni:
We use psr/http-message interfaces but don't implement them. So looks like our code is already compatible and we can switch to the new version smoothly.
As a bonus, fixed some deprecations.


e0ipso:
Thans @dzmitrykruchko-nbcuni !!

PR #202, opened 2025-10-02
chore: apply CS Fixer

e0ipso:
I executed: `vendor/bin/php-cs-fixer fix --config=.php-cs-fixer.dist.php -v --using-cache=no --path-mode=intersection -- src tests` and committed the results.


marcoscano:
👍 looks good, thanks for working on this!

